PR #1, opened 2017-06-23
[GSoC] Added LSTM baseline model with copy task benchmark

17minutes:
This PR is part of my GSoC project "Augmented RNNs".
Implemented a basic LSTM model for `AddTask` from the running version of my `mlpack` fork.

### DIFF
--- a/augmented_lstm_baseline/src/lstm_baseline_main.cpp
+++ b/augmented_lstm_baseline/src/lstm_baseline_main.cpp
@@ -91,185 +91,22 @@ RNN<MeanSquaredError<> > GenerateModel(size_t inputSize,
   return model;
 }
 
-//! Run CopyTask on the specified sequence length and repeat count.
-void RunCopyTask(size_t maxLen,
-                 size_t nRepeats,
-                 size_t epochs,
-                 size_t samples)
+//! Runs an instance of the task of given size on a baseline LSTM model.
+template<typename TaskType>
+void RunTask(TaskType& task,
+             size_t inputSize,
+             size_t outputSize,
+             size_t epochs,
+             size_t samples)
 {
-  Log::Info << "Running copy task benchmark for nRepeats = "
-             << nRepeats << ", maxLen = " << maxLen << "\n";
-
-  const size_t outputSize = 1;
-  const size_t inputSize = 2;
-  size_t maxRho = inputSize * (nRepeats + 1) * maxLen + 1;
+  // TODO Make this dirty hack less dirty.
+  size_t maxRho = inputSize * 1024 + 1;
 
   // Creating a baseline model.
   RNN<MeanSquaredError<> > model = GenerateModel(inputSize, outputSize, maxRho);
+  Adam opt;
 
-  Adam<decltype(model)> opt(model);
-
-  // Initializing task instance.
-  CopyTask task(maxLen, nRepeats, true);
-
-  arma::mat trainPredictor, trainResponse;
-  task.Generate(trainPredictor, trainResponse, samples);
-
-  arma::field<arma::mat> testPredictor, testResponse;
-  task.Generate(testPredictor, testResponse, samples, true);
-  assert(testPredictor.n_elem == testResponse.n_elem &&
-         testResponse.n_elem == samples);
-
-  // Training loop
-  Log::Info << "Running training loop for " << epochs << " epochs.\n";
-  for (size_t epoch = 0; epoch < epochs; ++epoch) {
-    Log::Debug << "Starting training epoch #"
-               << epoch+1 << "\n";
-    model.Rho() = trainPredictor.n_rows / inputSize;
-    model.Train(trainPredictor, trainResponse, opt);
-    Log::Debug << "Finished running training epoch #"
-               << epoch+1 << "\n";
-    std::cerr  << "Finished running training epoch #"
-               << epoch+1 << "\n";
-  }
-  Log::Info << "Finished training loop.\n";
-
-  // Evaluation loop
-  Log::Info << "Running evaluation loop.\n";
-  arma::field<arma::mat> modelOutput(samples);
-
-  for (size_t example = 0; example < samples; ++example) {
-    arma::mat predictor = testPredictor.at(example);
-    arma::mat response = testResponse.at(example);
-
-    Log::Debug << "Evaluating model on:\n";
-    Log::Debug << "Input sequence:\n" << predictor.t();
-    Log::Debug << "Ground truth sequence:\n" << response.t();
-
-    model.Rho() = predictor.n_elem / inputSize;
-    arma::mat softOutput;
-    model.Predict(
-      predictor,
-      softOutput);
-
-    Log::Debug << "Model predictions:\n";
-    Log::Debug << softOutput.t();
-
-    modelOutput.at(example) = softOutput.rows(
-      predictor.n_elem / (inputSize * (nRepeats + 1)),
-      softOutput.n_rows-1);
-    testResponse.at(example) = testResponse.at(example).rows(
-      predictor.n_elem / (inputSize * (nRepeats + 1)),
-      softOutput.n_rows-1);
-    Binarize<double>(modelOutput.at(example), modelOutput.at(example), 0.5);
-
-    Log::Debug << "Model predictions after binarization:\n";
-    Log::Debug << testResponse.at(example).t();
-  }
-  Log::Info << "Final score: "
-            << SequencePrecision<arma::mat>(testResponse, modelOutput)
-            << "\n";
-  std::cout << "Final score: "
-            << SequencePrecision<arma::mat>(testResponse, modelOutput)
-            << "\n";
-}
-
-
-//! Run AddTask on the specified bit length.
-void RunAddTask(size_t bitLen,
-                size_t epochs,
-                size_t samples)
-{
-  Log::Info << "Running add task benchmark for bitLen = " << bitLen << "\n";
-
-  const size_t outputSize = 3;
-  const size_t inputSize = 3;
-  size_t maxRho = inputSize * (bitLen + 1) + 1;
-
-  // Creating a baseline model.
-  RNN<MeanSquaredError<> > model = GenerateModel(inputSize, outputSize, maxRho);
-
-  Adam<decltype(model)> opt(model);
-
-  // Initializing task instance.
-  AddTask task(bitLen);
-
-  arma::mat trainPredictor, trainResponse;
-  task.Generate(trainPredictor, trainResponse, samples);
-
-  arma::field<arma::mat> testPredictor, testResponse;
-  task.Generate(testPredictor, testResponse, samples, true);
-  assert(testPredictor.n_elem == testResponse.n_elem &&
-         testResponse.n_elem == samples);
-
-  // Training loop
-  Log::Info << "Running training loop for " << epochs << " epochs.\n";
-  for (size_t epoch = 0; epoch < epochs; ++epoch) {
-    Log::Debug << "Starting training epoch #"
-               << epoch+1 << "\n";
-    model.Rho() = trainPredictor.n_rows / inputSize;
-    model.Train(trainPredictor, trainResponse, opt);
-    Log::Debug << "Finished running training epoch #"
-               << epoch+1 << "\n";
-    std::cerr  << "Finished running training epoch #"
-               << epoch+1 << "\n";
-  }
-  Log::Info << "Finished training loop.\n";
-
-  // Evaluation loop
-  Log::Info << "Running evaluation loop.\n";
-  arma::field<arma::mat> modelOutput(samples);
-  for (size_t example = 0; example < samples; ++example) {
-    arma::mat predictor = testPredictor.at(example);
-    arma::mat response = testResponse.at(example);
-
-    Log::Debug << "Evaluating model on:\n";
-    Log::Debug << "Input sequence:\n" << predictor.t();
-    Log::Debug << "Ground truth sequence:\n" << response.t();
-
-    model.Rho() = predictor.n_elem / inputSize;
-    arma::mat softOutput;
-    model.Predict(
-      predictor,
-      softOutput);
-
-    Log::Debug << "Model predictions:\n";
-    Log::Debug << softOutput.t();
-
-    modelOutput.at(example) = softOutput;
-    Binarize<double>(modelOutput.at(example), modelOutput.at(example), 0.5);
-
-    Log::Debug << "Model predictions after binarization:\n";
-    Log::Debug << testResponse.at(example).t();
-  }
-  Log::Info << "Final score: "
-            << SequencePrecision<arma::mat>(testResponse, modelOutput)
-            << "\n";
-  std::cout << "Final score: "
-            << SequencePrecision<arma::mat>(testResponse, modelOutput)
-            << "\n";
-}
-
-//! Run SortTask on the specified sequence length and bit length.
-void RunSortTask(size_t maxLen,
-                 size_t bitLen,
-                 size_t epochs,
-                 size_t samples) {
-  Log::Info << "Running sort task benchmark for maxLen = "
-            << maxLen << "\n, bitLen = " << bitLen << "\n";
-
-  size_t outputSize = bitLen;
-  size_t inputSize = bitLen;
-  size_t maxRho = inputSize * (bitLen + 1) + 1;
-
-  // Creating a baseline model.
-  RNN<MeanSquaredError<> > model = GenerateModel(inputSize, outputSize, maxRho);
-
-  Adam<decltype(model)> opt(model);
-
-  // Initializing task instance.
-  SortTask task(maxLen, bitLen);
-
+  // Generating a task instance
   arma::mat trainPredictor, trainResponse;
   task.Generate(trainPredictor, trainResponse, samples);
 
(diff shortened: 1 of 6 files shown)


partobs-mdp:
By the way, how can I suppress `SGD` output in log and at the same time see my `[DEBUG]` output?

zoq:
Thanks for the PR. Regarding the logging you could set:

```
mlpack::Log::Debug.ignoreInput = false;
mlpack::Log::Info.ignoreInput = true;
mlpack::Log::Warn.ignoreInput = true;
```

partobs-mdp:
Can't compile it:
```
/home/pretorium/Experiments/mlpack-models/models/augmented_lstm_baseline/src/lstm_baseline_main.cpp:276:14: error: ‘class mlpack::util::NullOutStream’ has no member named ‘ignoreInput’
   Log::Debug.ignoreInput = false;
```

zoq:
Right this only works if you build with ``DEBUG=ON``, but it looks like you build with ``DEBUG=OFF``?

partobs-mdp:
About logging: with `DEBUG=ON` this didn't work either.

About the last commit: trying to make an AddTask solver, but something goes wrong in a very weird way. I have already found an error (the sequences must be aligned by length), fixing.

partobs-mdp:
Well, I fixed it faster than I thought :smile:

Speaking about the LSTM model itself, it performs real poor on `AddTask` - it scored ~14.5% on the task of adding 2-bit numbers.

partobs-mdp:
I think I'll also create a a TODO list similar to the one from main `mlpack` repo:

- [ ] Create a function for creating a model (so far we have a copy-paste);
- [ ] Shuffle the examples in the train run;
- [ ] Evaluate the results and write them to the report.

zoq:
I'll take a look at the AddTask results, in my tested for the Copy Task, I usually use SGD or Mini Batch SGD to achieve better results. Can you push the Add and Sort task to the mlpack repo?

partobs-mdp:
> Can you push the Add and Sort task to the mlpack repo?

Sorry, didn't quite understand that - those already seem to be in `partobs-mdp/mlpack`. Can you restate that, please?

zoq:
> Thanks for the clarification, I guess relaying on a binary representation might be helpful if you are going to normalise the data later. So I think the next step is to transform the output into the representation as done for the copy task? But instead of two inputs we use 3 inputs?

> Yes, that's precisely what I'm trying to do now. The transform shouldn't be too hard - inputs.reshape, maybe preceeded with inputs.t()

I can't see that you changed the representation of the Sort or Add task, maybe I missed something?

zoq:
Ahh I see, you did the reshape inside the main file, I think it's a good idea to do this inside the task itself. What do you think?

partobs-mdp:
> Ahh I see, you did the reshape inside the main file, I think it's a good idea to do this inside the task itself. What do you think?

Sure, doing it right now.

rcurtin:
The bit about `Log::Debug` was backwards, we have to set `Log::Debug.ignoreInput = true` to suppress output.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,530 @@
...
+ * @author Konstantin Sidorov
+ *
+ * Executable for LSTM baseline solution for ann::augmented tasks.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <iostream>
```
rcurtin: This should be in prereqs.hpp, so you could remove this.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,530 @@
...
+      Log::Fatal << "Invalid value for 'samples': "
+                 << "expecting a positive number, received "
+                 << samples << "\n";
+      okParams = false;
+    }
+    if (okParams)
+      RunSortTask(maxLen, bitLen, epochs, samples);
+  }
+  else {
+    Log::Fatal << "Can't recognize task type, aborting execution.\n";
```
rcurtin: It would be nice to put the valid choices of task in this error message, so that the user doesn't then have to run with `-h` to fix their error.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,530 @@
...
+      Log::Fatal << "Invalid value for 'length': "
+                 << "expecting a positive number, received "
+                 << maxLen << "\n";
+      okParams = false;
+    }
+    int epochs = CLI::GetParam<int>("epochs");
+    if (repeats <= 0) {
+      Log::Fatal << "Invalid value for 'epochs': "
+                 << "expecting a positive number, received "
+                 << repeats << "\n";
```
rcurtin: There's a lot of repeated code here, I'd suggest refactoring it into the first part so it can exist only once.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,530 @@
...
+      RunCopyTask(maxLen, repeats, epochs, samples);
+  }
+  else if (task == "add") {
+    bool okParams = true;
+    int bitLen = CLI::GetParam<int>("bit_length");
+    if (bitLen <= 0) {
+      Log::Fatal << "Invalid value for 'bit_length': "
+                 << "expecting a positive number, received "
+                 << bitLen << "\n";
+      okParams = false;
```
rcurtin: Log::Fatal terminates the program so you don't need to worry about anything after you send a newline character to it.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,530 @@
...
+}
+
+//! Run SortTask on the specified sequence length and bit length.
+void RunSortTask(size_t maxLen,
+                 size_t bitLen,
+                 size_t epochs,
+                 size_t samples)
+{
+  Log::Info << "Running sort task benchmark for maxLen = "
+            << maxLen << "\n, bitLen = " << bitLen << "\n";
```
rcurtin: Are both new lines supposed to be here? The first looks accidental.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,530 @@
...
+  std::cout << "Final score: "
+            << SequencePrecision<arma::mat>(testResponse, modelOutput)
+            << "\n";
+}
+
+//! Run SortTask on the specified sequence length and bit length.
+void RunSortTask(size_t maxLen,
+                 size_t bitLen,
+                 size_t epochs,
+                 size_t samples)
```
rcurtin: Is this similar enough to `RunAddTask` that you could templatize it and combine the code?

partobs-mdp: True, but don't they have different parameters (I mean, `AddTask` has no `maxLen` - can I do some nice templating even though parameter sets are different?)

augmented_lstm_baseline/src/lstm_baseline_main.cpp
partobs-mdp: That's true, but I'm not 100% sure I can code it up today. So, I'll write up the idea so that your comment won't get collapsed after my commit: creating a list of params, iterating over it, saving a variable value to `vector`. If it is `<0`, print a crash message.

rcurtin: Ah, yes, I see that not all of the parameters are used by all of the methods, but still it would be easy to collapse down some of them.  Another problem with repeated code is that it usually gets copy-pasted and sometimes errors persist, like here (which I did not notice yesterday): `if (repeats <= 0)` should be `if (epochs <= 0)`. :)

Another strategy would just be to cast to a size_t.  I usually would do that: `const size_t epochs = (size_t) CLI::GetParam<int>("epochs");`.  Maybe the best idea of all would be to make a `PARAM_SIZE_T()` macro...

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,208 @@
...
+  task.Generate(trainPredictor, trainResponse, samples);
+
+  arma::field<arma::mat> testPredictor, testResponse;
+  task.Generate(testPredictor, testResponse, samples, true);
+  assert(testPredictor.n_elem == testResponse.n_elem &&
+         testResponse.n_elem == samples);
+
+  // Training loop
+  Log::Info << "Running training loop for " << epochs << " epochs.\n";
+  for (size_t epoch = 0; epoch < epochs; ++epoch) {
```
rcurtin: A high-level comment on this training loop---typically one epoch of training refers to a single pass over the dataset. But in this case the code is actually training a model using the default maximum number of iterations (this is 100000/numPoints epochs), then throwing the model away and retraining from scratch `epochs` times.

So instead, you should modify the `opt` object to set `MaxIterations()` equal to `epochs * numPoints`.  This will make the whole thing go a lot faster too. :)  Possibly, you may also want to set the tolerance of the optimizer to `0`, so it does not terminate early if it has converged.  That would be useful if you want the user to always expect that if they say `--epochs 100` then 100 is exactly the number of epochs that have run.  For `--samples=100`, I found that Adam would converge to the default tolerance usually in about 80-110 epochs for the copy task.

The SGD (and Adam) objects also print to `Log::Info` regularly, so there is no need to print any status output during training.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,208 @@
...
+    AddTask task(bitLen);
+    RunTask<AddTask>(task, 3, 3, epochs, samples);
+  }
+  else if (task == "sort") {
+    SortTask task(maxLen, bitLen);
+    RunTask<SortTask>(task, bitLen, bitLen, epochs, samples);
+  }
+  else {
+    Log::Fatal << "Can't recognize task type, aborting execution.\n";
+    Log::Fatal << "Supported tasks: add, copy, sort.\n";
```
rcurtin: A tricky bit about `Log::Fatal` is that it throws an exception once it receives a newline, so you should put these both on one line.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,208 @@
...
+    make_pair("bit_length", bitLen),
+    make_pair("length", maxLen),
+    make_pair("epochs", epochs),
+    make_pair("samples", samples)
+  };
+  for (pair<string, int> param : params) {
+    if (param.second <= 0) {   
+      Log::Fatal << "Invalid value for '" << param.first << "': "
+                 << "expecting a positive number, received "
+                 << param.second << "\n";
```
rcurtin: Can you add a period at the end of the sentence?  I know it is a pedantic comment. :)

augmented_lstm_baseline/src/run_add_task.sh
```diff
@@ -0,0 +1,5 @@
+for i in $(seq 1 10); do
+    echo "Evaluating bit_length $i"
+    COMMAND="./mlpack_lstm_baseline -t add -i 500 -b $i -r 1 -s 128 2> /dev/null | tail -n 1";
+    eval "$COMMAND";
+done
```
rcurtin: Can you add the prefix

```
#!/bin/bash
#
# <some comment about what the program does>
```

to these scripts?  Also, what directory should these be run from?  You could always take a command-line parameter for what the binary directory that `mlpack_lstm_baseline` is in.  Alternately, we could leave these scripts as-is and just mention in a comment that they would need manual adaptation for usage.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,208 @@
...
+    "combination of those options (as in case with --task=sort)."
+    "\n\n"
+    "Learning-specific parameters are also tunable (e.g., --epochs and --samples)");
+
+PARAM_STRING_IN_REQ("task", "Task to execute LSTM network on.", "t");
+
+PARAM_INT_IN("length", "Maximum sequence length (doesn't impact binary addition task).", "l", 2);
+PARAM_INT_IN("repeats", "Number of repeats required to solve the task (specific for Copy task).", "r", 1);
+PARAM_INT_IN("bit_length", "Maximum length of sequence elements in binary representation.", "b", 2);
+PARAM_INT_IN("epochs", "Learning epochs.", "i", 25);
```
rcurtin: Why not `-e` for epochs?

partobs-mdp:
> Did I run any of those simulations wrong, or is this the expected performance?

Everything goes as planned :) Neither me nor @zoq (as far as I remember) didn't manage to get outstanding performance of LSTM model on those two tasks.

rcurtin:
Ok, good to hear I did not encounter any issues that weren't expected.  If the performance is not so good, I wonder if the fix about the epochs I mentioned will allow it to train better---like if we use a very large number of epochs or something, as opposed to the current limit of 100k SGD iterations.  If you fix the code, I am happy to fire-and-forget run it in the background and let you know what the results are in a few days. :)

augmented_lstm_baseline/src/lstm_baseline_main.cpp
zoq: The problem I see is, that we train ``MaxIterations`` times on a single sample, since right now there is no interface for ``arma::field``; it should be easy to implement such an interface, or we could just use the fixed sequence representation and use ``arma::mat`` as input. 

Using ``arma::field`` the way we do it right now introduces another problem, at the beginning of each ``Optimize(...)`` call, we instantiate a new Adam update policy, which should be used over iterations.

augmented_lstm_baseline/src/lstm_baseline_main.cpp
```diff
@@ -0,0 +1,208 @@
...
+  int maxLen = CLI::GetParam<int>("length");
+  int epochs = CLI::GetParam<int>("epochs");
+  int samples = CLI::GetParam<int>("samples");
+  vector<pair<string, int>> params = {
+    make_pair("repeats", repeats),
+    make_pair("bit_length", bitLen),
+    make_pair("length", maxLen),
+    make_pair("epochs", epochs),
+    make_pair("samples", samples)
+  };
```
partobs-mdp: Sorry, didn't quite understand you - could you rephrase your comment?

zoq: "When wrapping a line, the next line should be tabbed twice from where the previous line began. An example is easier to understand:"  https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping

partobs-mdp:
Did I wrap the line correctly?

zoq:
> Did I wrap the line correctly?

Yes, looks good. I have to admit that we are really picky about style issues, sorry.

partobs-mdp:
> I have to admit that we are really picky about style issues, sorry.

I understand, no problem - otherwise the codebase would already degenerate into something completely unmaintainable and downright dirty :)

zoq:
Here are the updated results:

```
➜  bin git:(master) ✗ ./run_sort_task.sh 
Evaluating length = 1, binary_length = 2
Evaluating length = 2, binary_length = 2
Final score: 0.578125
Evaluating length = 3, binary_length = 2
Final score: 0.273438
Evaluating length = 4, binary_length = 2
Final score: 0.226562
Evaluating length = 5, binary_length = 2
Final score: 0.1875
Evaluating length = 6, binary_length = 2
Final score: 0.0859375
Evaluating length = 7, binary_length = 2
Final score: 0.046875
Evaluating length = 8, binary_length = 2
Final score: 0.0390625
Evaluating length = 1, binary_length = 3
Evaluating length = 2, binary_length = 3
Final score: 0.445312
Evaluating length = 3, binary_length = 3
Final score: 0.132812
Evaluating length = 4, binary_length = 3
Final score: 0.015625
Evaluating length = 5, binary_length = 3
Final score: 0.0078125
Evaluating length = 1, binary_length = 4
Evaluating length = 2, binary_length = 4
Final score: 0.320312
Evaluating length = 3, binary_length = 4
Final score: 0.03125
Evaluating length = 4, binary_length = 4
Final score: 0.0234375
Evaluating length = 1, binary_length = 5
Evaluating length = 2, binary_length = 5
Final score: 0.21875
Evaluating length = 3, binary_length = 5
Final score: 0
Evaluating length = 1, binary_length = 6
Evaluating length = 2, binary_length = 6
Final score: 0.0546875
Evaluating length = 1, binary_length = 7
Evaluating length = 2, binary_length = 7
Final score: 0.03125
Evaluating length = 1, binary_length = 8
Evaluating length = 2, binary_length = 8
Final score: 0.0234375
Evaluating length = 1, binary_length = 9
Evaluating length = 1, binary_length = 10
```

```
➜  bin git:(master) ✗ ./run_copy_task.sh  
Evaluating length = 2
Final score: 1
Evaluating length = 3
Final score: 1
Evaluating length = 4
Final score: 1
Evaluating length = 5
Final score: 1
Evaluating length = 6
Final score: 1
Evaluating length = 7
Final score: 0.96875
Evaluating length = 8
Final score: 0.421875
Evaluating length = 9
Final score: 0.84375
Evaluating length = 10
Final score: 0.820312
```

```
➜  bin git:(master) ✗ ./run_add_task.sh 
Evaluating bit_length 1
Final score: 0.460938
Evaluating bit_length 2
Final score: 0.28125
Evaluating bit_length 3
Final score: 0.109375
Evaluating bit_length 4
Final score: 0.0390625
Evaluating bit_length 5
Final score: 0.0234375
Evaluating bit_length 6
Final score: 0.0234375
Evaluating bit_length 7
Final score: 0
Evaluating bit_length 8
Final score: 0.0078125
Evaluating bit_length 9
Final score: 0
Evaluating bit_length 10
Final score: 0
```

rcurtin:
To add on to the discussion about training:

> The problem I see is, that we train MaxIterations times on a single sample, since right now there is no interface for arma::field; it should be easy to implement such an interface, or we could just use the fixed sequence representation and use arma::mat as input.
>
> Using arma::field the way we do it right now introduces another problem, at the beginning of each Optimize(...) call, we instantiate a new Adam update policy, which should be used over iterations.

As the code stands right now, it is using the fixed-length representation, but the testing is done on the non-fixed-length representation.  I would suspect that this might cause some decrease in performance, because the model always "expects" the sequence to be the same length as its training sequences (I am handwaving but that hopefully kind of makes sense).  I think maybe there are also some other issues, noted in mlpack#1005 about the representation of the sequence, but I do think this may play into it a little bit.

But, currently there is no way to pass non-fixed-length data.  I seem to remember having this discussion in another issue or PR, but I can't find it now.  @zoq, what do you think, should the `RNN::Train()` interface be changed to accept `arma::field<arma::mat>`, where each `arma::mat` is a sequence to train on?  I can't think of any other good solution to accept a set of non-fixed-length sequence inputs.  (That would then mean we should change the HMM interface, which uses `std::vector<arma::mat>` to accept a bunch of sequences, but of course that is no problem.)

zoq:
> But, currently there is no way to pass non-fixed-length data. I seem to remember having this discussion in another issue or PR, but I can't find it now. @zoq, what do you think, should the RNN::Train() interface be changed to accept arma::field<arma::mat>, where each arma::mat is a sequence to train on? I can't think of any other good solution to accept a set of non-fixed-length sequence inputs. (That would then mean we should change the HMM interface, which uses std::vector<arma::mat> to accept a bunch of sequences, but of course that is no problem.)

I think we should provide a Train interface for ``arma::mat`` and ``arma::field``, in case someone like sto use the same data as for training e.g. a feedforward network, we don't have to transform the data into ``arma::field``. Also, I don't mind to use ``std::vector<arma::mat>``instead of ``arma::filed``,  however for me ``arma::filed`` feels more natural, and I think has some additional benefits like ``.n_elem``.

rcurtin:
Agreed---`arma::mat` can assume a fixed-length representation, and `arma::field` can assume a non-fixed length representation.  I also agree that `arma::field<arma::mat>` is more natural than `std::vector<arma::mat>`.

sumedhghaisas:
I ran some experiments and found out that RMSProp performs the best. The learning rate I used was 0.0001 with default momentum. The learning for LSTM was much more stable and much faster. Each iteration with RMSProp is 3 times faster than that of Minibatch SGD (on my machine). Same for NTM.

rcurtin:
Sumedh, thanks for running the experiments.

@partobs-mdp: what would you say the status of this is?  I realize I may have missed some IRC discussions while I was traveling.  Does this need to be further reviewed, or is it waiting on some more implementation from your end, or is it ready to merge?

partobs-mdp:
@rcurtin: It's mostly finished, but @zoq (if I remember correctly) has promised to run LSTM model on "shifted" representation of AddTask and SortTask and show the results.

zoq:
First results with the shifted version:

```
➜  bin git:(master) ✗ ./run_sort_task.sh 
Evaluating length = 1, binary_length = 2
Evaluating length = 2, binary_length = 2
Average score: 1
Evaluating length = 3, binary_length = 2
Average score: 1
Evaluating length = 4, binary_length = 2
Average score: 0.851562
Evaluating length = 5, binary_length = 2
Average score: 0.929688
Evaluating length = 6, binary_length = 2
Average score: 0.648438
Evaluating length = 7, binary_length = 2
Average score: 0.671875
Evaluating length = 8, binary_length = 2
Average score: 0.671875
Evaluating length = 1, binary_length = 3
Evaluating length = 2, binary_length = 3
Average score: 0.945312
Evaluating length = 3, binary_length = 3
Average score: 0.601562
Evaluating length = 4, binary_length = 3
Average score: 0.234375
Evaluating length = 5, binary_length = 3
Average score: 0.148438
Evaluating length = 1, binary_length = 4
Evaluating length = 2, binary_length = 4
Average score: 0.710938
Evaluating length = 3, binary_length = 4
Average score: 0.0703125
Evaluating length = 4, binary_length = 4
Average score: 0
Evaluating length = 1, binary_length = 5
Evaluating length = 2, binary_length = 5
Average score: 0.21875
Evaluating length = 3, binary_length = 5
Average score: 0.015625
Evaluating length = 1, binary_length = 6
Evaluating length = 2, binary_length = 6
Average score: 0.078125
Evaluating length = 1, binary_length = 7
Evaluating length = 2, binary_length = 7
Average score: 0
Evaluating length = 1, binary_length = 8
Evaluating length = 2, binary_length = 8
Average score: 0.0078125
Evaluating length = 1, binary_length = 9
Evaluating length = 1, binary_length = 10
```

I think if we adjust some network parameter we can improve the results.

partobs-mdp:
@zoq: Imho we don't have to adjust the model - our goal was to prove our task class and model are not broken.

zoq:
For completeness:
```
➜  bin git:(master) ✗ ./run_add_task.sh
Evaluating bit_length 1
Average score: 1
Evaluating bit_length 2
Average score: 1
Evaluating bit_length 3
Average score: 0.96875
Evaluating bit_length 4
Average score: 0.484375
Evaluating bit_length 5
Average score: 0.0078125
Evaluating bit_length 6
Average score: 0.015625
Evaluating bit_length 7
Average score: 0.0078125
Evaluating bit_length 8
Average score: 0
Evaluating bit_length 9
Average score: 0
Evaluating bit_length 10
Average score: 0
```

I agree, searching for optimal parameter isn't part of this project, I think https://github.com/mlpack/mlpack/pull/1005 is ready to be merged.

PR #14, opened 2018-07-05
[WIP] Add Variational Autoencoder model for MNIST

akhandait:
@sumedhghaisas


vae/vae_mnist/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ggdb3")
+else()
+  message("Compilation without debug info (without ggdb3 flag)")
+endif()
+
+add_executable(vae ${SOURCES})
+target_link_libraries(vae
+  ${MLPACK_LIBRARIES}
+  ${ARMADILLO_LIBRARIES}
+  ${Boost_Serialization_LIBRARIES})
```
zoq: Can you use ``${Boost_LIBRARIES}`` instead of ``${Boost_Serialization_LIBRARIES}``, I think ``${Boost_Serialization_LIBRARIES}`` is not defined.

vae/vae_generate.cpp
```diff
@@ -0,0 +1,59 @@
...
+  // Load the trained model.
+  FFN<MeanSquaredError<>, HeInitialization> vaeModel;
+  data::Load("saved_models/vaeModelCNN.xml", "vaeModelCNN", vaeModel);
+
+  arma::mat outputDists;
+
+  // Generate randomly from the trained distribution.
+  arma::mat samples = arma::randn<arma::mat>(20, n_samples);
+
+  vaeModel.Forward(samples, outputDists, 3, 3);
```
sumedhghaisas: This is confusing to read as it does not show clearly which kind of sampling is done, latent prior or latent posterior. The new Forward function symantics is unusual. Thus here we should have enough documentation to explain the user what is happening as models could be read by a normal user who is not aware of this semantics. Also we should show both sampling methods here.

vae/vae_generate.cpp
```diff
@@ -0,0 +1,59 @@
...
+    data::Split(fullData, validation, train, 0.8);
+  }
+
+  // Load the trained model.
+  FFN<MeanSquaredError<>, HeInitialization> vaeModel;
+  data::Load("saved_models/vaeModelCNN.xml", "vaeModelCNN", vaeModel);
+
+  arma::mat outputDists;
+
+  // Generate randomly from the trained distribution.
```
sumedhghaisas: I think here you sampling from latent prior which is N(0, 1) not the trained distribution.

vae/vae_generate.cpp
```diff
@@ -0,0 +1,59 @@
...
+  arma::mat samples = arma::randn<arma::mat>(20, n_samples);
+
+  vaeModel.Forward(samples, outputDists, 3, 3);
+  arma::mat outputSamples = outputDists;
+
+  outputSamples = outputSamples / 2 + 0.5;
```
sumedhghaisas: Lets explain why we post-process samples. ( Cause we have modeled the pre-processed data. :) )

vae/vae_cnn_mnist/src/vae_cnn.cpp
```diff
@@ -0,0 +1,215 @@
...
+  for (int i = 0; i < cycles; i++)
+  {
+    // Train neural network. If this is the first iteration, weights are
+    // random, using current values as starting point otherwise.
+    vaeModel.Train(train, train, optimizer);
+
+    // Don't reset optimizer's parameters between cycles.
+    optimizer.ResetPolicy() = false;
+
+    std::cout << "Loss after cycle  " << i << " -> " <<
```
sumedhghaisas: 1 testing batch is usually very unstable for test loss computation. In practice, we should average over some number of batches, I usually use 100, but even less than that will do.

vae/vae_mnist/src/vae.cpp
```diff
@@ -0,0 +1,166 @@
...
+  {
+    // Train neural network. If this is the first iteration, weights are
+    // random, using current values as starting point otherwise.
+    vaeModel.Train(train, train, optimizer);
+
+    // Don't reset optimizer's parameters between cycles.
+    optimizer.ResetPolicy() = false;
+
+    std::cout << "Loss after cycle " << i << " -> " <<
+        vaeModel.Evaluate(train_test, train_test) << std::endl;
```
sumedhghaisas: Same comment again, lets average over some number of batches.

vae/vae_mnist/src/vae.cpp
```diff
@@ -0,0 +1,166 @@
+/**
+ * @file vae.cpp
+ * @author Atharva Khandait
+ *
+ * A Variational autoencoder(VAE) model to generate MNIST.
```
sumedhghaisas: Feed forward model

vae/vae_cnn_mnist/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+    COMMAND ${CMAKE_COMMAND} -E tar xvzf
+      ${CMAKE_CURRENT_SOURCE_DIR}/../../data/mnist_full.tar.xz
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/../)
+endif()
+
+if (NOT EXISTS ${PROJECT_BINARY_DIR}/../saved_models)
+  add_custom_command(TARGET vae_cnn POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E make_directory
+      ${PROJECT_BINARY_DIR}/../saved_models)
+endif()
```
sumedhghaisas: Add an extra line in the end. :)

vae/generate_images.py
```diff
@@ -0,0 +1,22 @@
...
+
+  latent_varying = np.concatenate((latent_varying, temp_image), axis=1)
+
+  temp_image = Image.fromarray(temp_image, 'L')
+  temp_image.save('generated_samples_3/image' + str(i) + '.jpg')
+
+latent_varying = Image.fromarray(latent_varying, 'L')
+latent_varying.save('generated_samples_3/image_combined' + '.jpg')
+
+print (str(outputSamples.shape[1]) + ' samples saved in generated_samples_3/.')
```
sumedhghaisas: Extra line. :)

vae/vae_generate.cpp
akhandait: Oh sorry, this was when I was using the posterior to generate. I forgot to remove this comment.

vae/vae_mnist/src/vae.cpp
akhandait: Doesn't feed forward also include convolutional? Should we say something like dense VAE model in this case?

akhandait:
@sumedhghaisas I will test that new convolutional architecture tonight and then upload it.

vae/vae_cnn_mnist/src/vae_cnn.cpp
```diff
@@ -0,0 +1,240 @@
...
+        MeanTestLoss<ReconModel>(vaeModel, train_test, 50) << std::endl;
+    std::cout << "Time taken for cycle -> " << float(clock() - begin_time) /
+        CLOCKS_PER_SEC << " seconds" << std::endl;
+
+    if (saveModel)
+    {
+      data::Save("vae/saved_models/vaeCNNBinary.xml", "vaeCNNBinary",
```
rcurtin: You may want to save as `.bin`, the XML models will be really huge. :)

vae/vae_generate.cpp
```diff
@@ -0,0 +1,102 @@
...
+  // Whether modelled on binary data.
+  constexpr bool isBinary = false;
+  // the latent size of the VAE model.
+  constexpr size_t latentSize = 20;
+
+  arma::mat fullData, train, validation;
+
+  if (loadData)
+  {
+    data::Load("../build/vae/mnist_full.csv", fullData, true, false);
```
rcurtin: Should we load from `vae/mnist_full.csv` instead of `../build/vae/mnist_full.csv`?  It seems like `vae_cnn.cpp` uses that path, so I would think maybe it's a good idea to use the same path here.

vae/vae_cnn_mnist/src/vae_cnn.cpp
akhandait: Yeah, I have been saving the XML models and they are huge!

vae/vae_generate.cpp
akhandait: Okay, I will do it.

vae/vae_mnist/src/vae.cpp
sumedhghaisas: Yeah I am not sure about the exact terminology here :( The best thing to to do is use 90s terminology, lets say MLP VAE, multi-layer perception. No confusion with convolution then.

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -154,33 +153,33 @@ int main()
...
   // Saving results into Kaggle compatibe CSV file.
-  save("results.csv", "ImageId,Label", testPred);
-  cout << "Results were saved to \"results.csv\" and could be uploaded to " 
-    << "https://www.kaggle.com/c/digit-recognizer/submissions for a competition" 
+  save("Kaggle/results.csv", "ImageId,Label", testPred);
+  cout << "Results were saved to \"results.csv\" and could be uploaded to "
+    << "https://www.kaggle.com/c/digit-recognizer/submissions for a competition"
     << endl;
   cout << "Finished" << endl;
 }
```
sumedhghaisas: Extra line here. :)

sumedhghaisas:
I have posted some minor comments, but except for that the PR looks good to me. Lets we have to wait till we merge ReconstrutionLoss right?

vae/vae_generate.cpp
sumedhghaisas: I think I will agree with Ryan here. You don't have to build it, but we can make a cmake target to copy it to build folder.

akhandait: Done!

sumedhghaisas:
PR seems merge ready now. I will go ahead and merge this in 2 days in case someone wants to take another look. :)

CMakeLists.txt
```diff
@@ -167,7 +173,8 @@ set(MODELS_INCLUDE_DIRS ${MLPACK_INCLUDE_DIR}
                         ${Boost_INCLUDE_DIRS})
 set(MODELS_LIBRARIES ${MLPACK_LIBRARY}
                      ${ARMADILLO_LIBRARIES}
-                     ${Boost_LIBRARIES})
+                     ${Boost_LIBRARIES}
+                     ${OpenMP_LIBRARY})
```
zoq: Look like ``OpenMP_LIBRARY`` isn't necessary or defined: https://cmake.org/cmake/help/v3.0/module/FindOpenMP.html

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -154,33 +153,33 @@ int main()
...
     tempDataset.n_rows - 1, tempDataset.n_cols - 1);
 
   mat testPredOut;
   // Getting predictions on test data points .
   model.Predict(testX, testPredOut);
   // Generating labels for the test dataset.
   Row<size_t> testPred = getLabels(testPredOut);
-  cout << "Saving predicted labels to \"results.csv\" ..." << endl;
-  
+  cout << "Saving predicted labels to \"Kaggle/results.csv\" ..." << endl;
```
zoq: Thanks for updating the file.

Kaggle/DigitRecognizerCNN/src/DigitRecognizerCNN.cpp
```diff
@@ -47,7 +47,7 @@ int main()
   constexpr double STEP_SIZE = 1.2e-3;
 
   // Number of data points in each iteration of SGD.
-  constexpr int BATCH_SIZE = 1;
+  constexpr int BATCH_SIZE = 50;
```
zoq: I guess you increased the value to accelerate the training process.

vae/generate_images.py
```diff
@@ -0,0 +1,78 @@
...
+def ImagesFromCSV(filename,
+                  imgShape = (28, 28),
+                  destination = 'samples',
+                  saveIndividual = False):
+
+  # Import the data into a numpy matrix.
+  samples = np.genfromtxt(filename, delimiter=',', dtype=np.uint8)
+
+  # Reshape and save it as an image in the destination.
+  temp_image = Image.fromarray(np.reshape(samples[:, 0], imgShape), 'L')
```
zoq: I think we should use the same code style in the same file, so either we use ``temp_image`` or ``tempImage``, in the next line you used ``saveIndividual``.

vae/generate_images.py
```diff
@@ -0,0 +1,78 @@
...
+  if saveIndividual:
+    temp_image.save(destination + '/sample0.jpg')
+
+  # All the images will be concatenated to this for a combined image.
+  allSamples = temp_image
+
+  for i in range(1, samples.shape[1]):
+
```
zoq: Looks like we could remove the extra line here.

vae/generate_images.py
```diff
@@ -0,0 +1,78 @@
...
+# Save prior samples with individual latent varying.
+latentSize = 10
+allLatent = ImagesFromCSV('samples_csv_files/samples_prior_latent0.csv',
+    destination = 'samples_prior')
+
+for i in range(1, latentSize):
+  allLatent = np.concatenate((allLatent,
+      (ImagesFromCSV('samples_csv_files/samples_prior_latent'+str(i)+'.csv',
+      destination = 'samples_prior'))), axis=0)
```
zoq: Pedantic style issue: can we use the same style throughout the file, so either ``axis=0`` or ``axis = 0``?

vae/generate_images.py
```diff
@@ -0,0 +1,78 @@
...
+saved.save('samples_prior/allLatent.jpg')
+
+# Save prior samples with 2d latent varying.
+nofSamples = 20
+allLatent = ImagesFromCSV('samples_csv_files/samples_prior_latent_2d0.csv',
+    destination = 'latent')
+
+for i in range(1, nofSamples):
+  allLatent = np.concatenate((allLatent,
+      (ImagesFromCSV('samples_csv_files/samples_prior_latent_2d'+str(i)+'.csv',
```
zoq: Another pedantic style issue: it looks like we use spaces between operations in the previous lines.

vae/vae_cnn_mnist/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+cmake_minimum_required(VERSION 2.6)
+project(vae-cnn-mnist)
+
+option(DEBUG "DEBUG" OFF)
+
+set(MODEL_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/src)
+include_directories("${CMAKE_CURRENT_SOURCE_DIR}/../../")
+
+set(SOURCES
+  ${MODEL_SOURCE_DIR}/vae_cnn.cpp
```
zoq: Does this fit in a single line?

vae/vae_cnn_mnist/src/vae_cnn.cpp
```diff
@@ -0,0 +1,240 @@
...
+    optimizer.ResetPolicy() = false;
+
+    std::cout << "Loss after cycle  " << i << " -> " <<
+        MeanTestLoss<MeanSModel>(vaeModel, train_test, 50) << std::endl;
+    std::cout << "Time taken for cycle -> " << float(clock() - begin_time) /
+        CLOCKS_PER_SEC << " seconds" << std::endl;
+
+    if (saveModel)
+    {
+      data::Save("vae/saved_models/vaeCNN.bin", "vaeCNN",
```
zoq: Do you think it will work as well if we save the model once at the end of the training process? I guess, if we like to save intermediate models as well, we could introduce another parameter. Depending on the model size, saving might take some extra time.

Kaggle/DigitRecognizerCNN/src/DigitRecognizerCNN.cpp
akhandait: Oh, I think I had done that for testing something for myself. I will change it back to what it was.

rcurtin: If it's faster, personally I see no need to change it back; the results shouldn't be that different either way.

vae/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```cpp
+arma::mat dataset;
+data::Load("dataset.csv", dataset, true);
+```
+
+Now, we will take the features in a seperate matrix. In Armadillo representation, the rows and columns will be interchanged.
+
+```cpp
+arma::mat data = dataset.submat(1, 0, train.n_rows - 1, train.n_cols - 1);
```
rcurtin: Is this readme complete?  I don't know if its only purpose was to point out that Armadillo matrices are column-major.  You could also point to the mlpack tutorial about this too if that's all you were looking for.

akhandait: Oh, it was supposed to be a tutorial about these models. The only part I have added until now is how to prepare a dataset. I guess I will remove this file from this PR and open a new PR which will have the complete tutorial.

akhandait:
@zoq @rcurtin @sumedhghaisas Can you please take a look at why travis is failing? The errors seem to start with `E: Failed to fetch`.

zoq:
Looks like a temporary issue with the connection; let's restart the build.

PR #21, opened 2019-02-26
use ensmallen instead of deprecated optimizer

Soonmok:
Fixed deprecated optimzer header file problem.


Soonmok:
I didn't know that there was already the pull requrest for this problem..

PR #22, opened 2019-03-05
added gan

Soonmok:
Added examples of gan network on mnist dataset.
Added examples of Generating image from gan generator.


Soonmok:
Change to use ensmallen to pass build test

zoq:
There is an open PR https://github.com/mlpack/models/pull/18, which should fix the build issue, I'll see if I can fix the error in the next days.

zoq:
Thanks to @rcurtin the build issue is fixed, see https://github.com/mlpack/models/pull/18 for more information.

Soonmok:
I will

PR #28, opened 2019-06-09
Added VGG19 Model for MNIST Dataset.

MuLx10:
VGG19 Model for MNIST.
It uses Kaggle MNIST dataset


zoq:
Nice, do you have some first results for the model?

MuLx10:
I don't have any results as of now. I ran for 1 cycle to check if the code was working and model was being saved correctly.

zoq:
I see, can you run the model for a couple of iterations or should I start the training process?

MuLx10:
> I see, can you run the model for a couple of iterations or should I start the training process?

I have started the training.

sumedhghaisas:
Excited to see the results. :) How long did 1 iteration took? Just for a ballpark measure

VGG19/vgg19_mnist/src/VGG19.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include <mlpack/core.hpp>
+#include <mlpack/prereqs.hpp>
+#include <mlpack/methods/ann/layer/layer.hpp>
+#include <mlpack/methods/ann/ffn.hpp>
+
+using namespace mlpack;
+using namespace mlpack::ann;
+
+class VGG19
```
zoq: We should add some more information about the model here, a reference to the paper would also be helpful.

VGG19/vgg19_mnist/src/VGG19.hpp
```diff
@@ -0,0 +1,82 @@
...
+{
+public:
+
+  VGG19(const size_t inputWidth,
```
zoq: A description for each parameter would be nice to have.

VGG19/vgg19_mnist/src/VGG19.hpp
```diff
@@ -0,0 +1,82 @@
...
+        const size_t inputHeight,
+        const size_t inputChannel,
+        const size_t numClasses,
+        const bool includeTop = true,
+        const std::string pooling = "max",
+        const std::string weights = "imagenet"
```
zoq: Not sure I get the meaning behind the parameter looks like it's unused.

VGG19/vgg19_mnist/src/VGG19.hpp
```diff
@@ -0,0 +1,82 @@
...
+        );
```
zoq: This should end right behind the laster parameter.

VGG19/vgg19_mnist/src/VGG19_impl.hpp
```diff
@@ -0,0 +1,182 @@
...
+#include "VGG19.hpp"
+
+VGG19::VGG19(const size_t inputWidth,
+             const size_t inputHeight,
+             const size_t inputChannel,
+             const size_t numClasses,
+             const bool includeTop,
+             const std::string pooling,
+             const std::string weights
+            ):
```
zoq: See comment above.

VGG19/vgg19_mnist/src/VGG19_impl.hpp
```diff
@@ -0,0 +1,182 @@
...
+        2,  // Width of field.
+        2,  // Height of field.
+        2,  // Stride along width.
+        2,  // Stride along height.
+        true
+      );
+
+    std::cout << "Width" << " " << "Height" << std::endl;
+
+    // same padding
```
zoq: Please use camel casing for all names.

VGG19/vgg19_mnist/src/VGG19.hpp
MuLx10: Pooling is for choosing between global max/avg pooling.
weights parameter would make the VGG load imagenet trained weights, else it would make VGG start from scratch. Since it is mnist based hence not used
.

zoq:
Any updates on the results?

VGG19/vgg19_mnist/src/VGG19_impl.hpp
```diff
@@ -0,0 +1,181 @@
...
+    }
+    return VGGNet;
+}
+
+size_t VGG19::GetOutputShape()
+{
+    return outputShape;
+}
+
+Sequential<>* VGG19::LoadModel(std::string filePath)
```
zoq: What about using ``const std::string& ``here and for the save method as well; avoids another copy.

MuLx10:
> Any updates on the results?

Four cycles completed.
Epoch 0
Loss after cycle 0 -> 311.907
Epoch 1
Loss after cycle 1 -> 311.83
Epoch 2
Loss after cycle 2 -> 311.731
Epoch 3
Loss after cycle 3 -> 311.658
Epoch 4
Loss after cycle 4 -> 311.349

zoq:
Hm, maybe it makes sense to start on a subset to figure out some reasonable optimizer settings, looking at the first iterations I would increase the step size.

sumedhghaisas:
What step size did you use for this training? I agree with Marcus here to increase the step size.

MuLx10:
Initial step size was 1.2e-3. I changed it to 5e-3 and restarted the training. Should I increase it more?

sumedhghaisas:
hmm... thats interesting. Usually the VGG models (smaller ones) I train require 1e-3 step size. I checked the Adam configuration you used and that should do it. For bigger models higher step size require higher batch size, or batch normalization. How does the rest of the training curve look like? Does the loss go down later?

sumedhghaisas:
Another thing you could try is using high learning rate in the start but decrease it later.

zoq:
I would also test another weight initialization policy.

zoq:
@MuLx10 Any update on this one?

MuLx10:
New paramters (lr = 5e-3) caused divergence.
Epoch 0
Loss after cycle 0 -> 310.261
Epoch 1
Loss after cycle 1 -> 223.393
Epoch 2
Loss after cycle 2 -> 2.83091e+10
Epoch 3
Loss after cycle 3 -> 6.03322e+45
Epoch 4
Loss after cycle 4 -> 1.6653e+49
Epoch 5
Loss after cycle 5 -> 3.37138e+23
Epoch 6
Loss after cycle 6 -> 8.95541e+53

VGG19/vgg19_mnist/src/VGG19_impl.hpp
```diff
@@ -0,0 +1,181 @@
...
+        // Add the final dense layer.
+        VGGNet->Add<Linear<> >(inputWidth * inputHeight * 512, 4096);
+        VGGNet->Add<Linear<> >(4096, 4096);
+        VGGNet->Add<Linear<> >(4096, numClasses);
+        outputShape = numClasses;
+    }
+    else
+    {   
+        outputShape = inputWidth * inputHeight * 512;
+        if (pooling == "max")
```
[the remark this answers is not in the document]
MuLx10: Done as you told.

zoq:
Okay, do you plan to test the ideas mentioned above next?

MuLx10:
I have restarted the training with different initialization and step size.

zoq:
Btw. if you don't see reasonable results in the first 10 iterations, you can stop the training process; in this case there might be something wrong with the model, in which case let's create a super simple subset (like 50 samples) and test on that one.

VGG19/vgg19_mnist/src/model.cpp
```diff
@@ -82,35 +83,34 @@ int main()
...
+  X /= 255.0;
+  // Create labels for the datasets.
+  data::OneHotEncoding(arma::conv_to<arma::irowvec>::from(dataset.row(0)), Y);
+  Y = Y.t();
 
-  // The train and valid datasets contain both - the features as well as the
-  // class labels. Split these into separate mats.
-  const mat trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1);
-  const mat validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1);
+  // Split the dataset into training and validation sets.
```
zoq: You could use: https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/data/split_data.hpp for the splitting as well.

MuLx10: Sure.

zoq:
I think once we see good results (is there any update on that point?), this is ready to be merged. A simple tutorial about the model would be nice to have.

MuLx10:
Sure I'll add it.

VGG19/vgg19_mnist/src/model.cpp
```diff
@@ -10,7 +10,7 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 
-#include "VGG19.hpp"
+#include "../../VGG19.hpp"
```
zoq: Nice catch!

MuLx10:
> Any update on the tutorial part?

I'll add tutorial for both (mnist & imagenet) the example together.

zoq:
Any update on that? Maybe CIFAR 10 is another solution that is way faster to train.

zoq:
Thanks for adding the tutorial, hopefully we see some good results.

PR #36, opened 2019-12-25
Rewritten and improved LSTM/TimeSeries-Multivariate AND LSTM/TimeSeries-Univariate tutorials

zsogitbe:
Rewritten and improved the tutorial. The data file Google2016-2019.csv was also not well sorted for the purpose of the task (also changed).


LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -9,9 +9,26 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  *
  * @file LSTMTimeSeriesMultivariate.cpp
- * @author Mehul Kumar Nirala
+ * @author Zoltan Somogyi based on the work of Mehul Kumar Nirala.
```
rcurtin: We can just keep Mehul's name also:

```
 * @author Mehul Kumar Nirala
 * @author Zoltan Somogyi
```

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -9,9 +9,26 @@
...
  */
 
+/*
+NOTE: the data need to be sorted by date in assending order! The RNN learns the past!
```
rcurtin:
```suggestion
NOTE: the data need to be sorted by date in ascending order! The RNN learns from oldest to newest!
```

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -30,159 +47,249 @@ using namespace ens;
...
+	// Getting predictions on test data points.
+	modelP.Predict(testX, predOutP);
+	// Calculating mse on prediction.
+	double testMSEP = MSE1(predOutP, testY);
+	cout << "Mean Squared Error on Prediction data points:= " << testMSEP << endl;
+
+	//save the output predictions
+	saveP(basedatapath + "/saved_models/LSTMMulti_predictions.csv", predOutP, scale, testX);
+
+	cout << "Ready!" << std::endl;
```
rcurtin: Why `Ready!`?  It seems like the program is done at this point and you could just exit.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -30,159 +47,249 @@ using namespace ens;
...
-  cout << "Finished" << endl;
-  cout << "Saving Model" << endl;
-  if (saveModel)
-  {
-    data::Save("saved_models/LSTMMulti.bin", "LSTMMulti", model);
-    std::cout << "Model saved in saved_models/." << std::endl;
-  }
-  return 0;
+	//TODO: change this to your base data path!
+	const string basedatapath = "D:/sample-ml-app";
```
rcurtin: I would suggest leaving this as something that's platform-independent.  This will never work for the majority of our users, who are using OS X or Linux.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -30,159 +47,249 @@ using namespace ens;
 /*
  * Function to calcute MSE for arma::cube
  */
-double MSE(arma::cube& pred, arma::cube& Y)
+double MSE1(arma::cube& pred, arma::cube& Y)
```
rcurtin: Why MSE1 and not just MSE?

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -30,159 +47,249 @@ using namespace ens;
...
+		X.subcube(span(), span(i), span()) = dataset.submat(span(), span(i, i + rho - 1));
+		y.subcube(span(), span(i), span()) = dataset.submat(span(3, 4), span(i + 1, i + rho));
+	}
+}
+
+/*
+ * This function saves the input data for prediction and the prediction results in CSV format. 
+ * The prediction results are the (high, low) for the next day and comming from the last slice 
+ * from the prediction.
+ */
```
rcurtin: Why don't we just use mlpack's existing `data::Save()` functionality for this?  I'm not sure why the need for something as complex as this.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: OK!

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: OK!

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: On Windows the console does not disappear because of the getchar function and the "Ready!" signals that the program is ready. This is needed on Windows. I can not help in multiplatform (unix, linux), please change the code if needed for multiplatform (sorry!).

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: This is needed on Windows. I can not help in multiplatform (unix, linux), please change the code if needed for multiplatform (sorry!).

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: OK! (I needed this because both tutorials use the same function name and it does not compile if both cpp files are present.)

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: > Hey @zsogitbe, thanks for putting the time into revising this tutorial. Could I ask that you fix the style so it matches the mlpack style guide before I review it further? With all the spacing changes, the diffs are very hard to read (i.e. it's hard for me to see what actually changed and what didn't). Thanks! I left a few comments so far. 👍

This has been done by Visual Studio no Windows. I tried to convert the tabs to 2 chars on the internet. This is the max I can do. I hope it is OK now. If not then please try to change it as needed (sorry!).

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: I find it important that we have a precise and nice output. The code saves the input and the two output columns in a CSV format. You can then just put it in Excel and make a chart from the results. As far as I know this is not possible with the data::Save().

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: There are several important changes which you should check out. The indexing of the input data is changed (see CreateTimeSeriesData). The sorting of the input data is completely changed (just the opposite). For all other things I have put a lot of comments in the code. The result is that the code is changed theoretically significantly!

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +61,114 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
- * Opening stock price, Closing stock price, Min stock price, Max stock price]
- * for past 'rho' days and the target variable will be Google’s
- * stock price today (high, low) and so on.
+ * The time series data for training the model contains the Closing stock price, the Volume of stocks traded,
+ * Opening stock price, Highest stock price and Lowest stock price for 'rho' days in the past. 
+ * The two target variables (multivariate) we want to predict are the Highest stock price and Lowest stock price
+ * (high, low) for the next day! 
+ *
+ * NOTE: Please note that we do not use the last input data point in the training because there is no target 
+ *       (next day (high, low)) for that point!
```
rcurtin: I'll go through during merge and make the lines 80 characters to match mlpack's style guide. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +61,114 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
  */
 template<typename InputDataType = arma::mat,
-     typename DataType = arma::cube,
-     typename LabelType = arma::cube>
-void CreateTimeSeriesData(InputDataType dataset, DataType& X, LabelType& y, size_t rho)
+  typename DataType = arma::cube,
+  typename LabelType = arma::cube>
+  void CreateTimeSeriesData(InputDataType dataset, DataType& X, LabelType& y, size_t rho)
```
rcurtin:
```suggestion
void CreateTimeSeriesData(InputDataType dataset, DataType& X, LabelType& y, size_t rho)
```
Seems like no need for extra spacing here. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: I think it's not too hard to simplify. :)  Here's a proposal (please double-check if it works---I haven't run it):

```
// We need the test data and predictions in the same format we scaled them in,
// so that we can unscale them to save the predictions.
mat flatDataAndPreds(testX.n_rows + 2, testX.n_cols);
flatDataAndPreds = testX.slice(testX.n_slices - 1);
flatDataAndPreds.rows(testX.n_rows, testX.n_rows + 1) = predictions.slice(predictions.n_slices - 1);

scale.InverseTransform(flatDataAndPreds);

// Save the data to file.  (The last column is the predictions; the preceding columns are
// the data used to generate those predictions.)
data::Save("predictions.csv", flatDataAndPreds);

// Print the output to screen.
std::cout << "The predicted Google stock (high, low) for the last day is: " << std::endl;
std::cout << "  (" << flatDataAndPreds(flatDataAndPreds.n_rows - 2, flatDataAndPreds.n_cols - 1)
std::cout << flatDataAndPreds(flatDataAndPreds.n_rows - 1, flatDataAndPreds.n_cols - 1) << ")" << std::endl;
```

I think that should work...

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +61,114 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
+  }
+
+  out.close();
+}
+
 int main()
 {
-  /* HYPERPARAMETERS */
+  //TODO: change this to your base data path!
+  const string basedatapath = "D:/sample-ml-app";
```
rcurtin: Actually I wonder if a nicer way to handle the fact that everyone may be on a different OS is to set all the filenames here.  Here's a suggestion:

```
// Change the names of these files as necessary.  They should be correct already, if your
// program's working directory contains the data and/or model.
const string dataFile = "Google2016-2019.csv";
// example: const string dataFile = "C:/mlpack-model-app/Google2016-2019.csv";
// example: const string dataFile = "/home/user/mlpack-model-app/Google2016-2019.csv";
const string modelFile = "lstm_multi.bin";
// example: const string modelFile = "C:/mlpack-model-app/lstm_multi.bin";
// example: const string modelFile = "/home/user/mlpack-model-app/lstm_multi.bin";
```

That way, a user can control all paths here.  We can keep it relative (since that will work on any platform... if the data is in the right place), and then if the user wants to change it they have some examples right there.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +61,114 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
-  // Number of iteration per epoch.
+  // Number of iterations per cycle.
   const int ITERATIONS_PER_EPOCH = 1000;
 
   // Step size of an optimizer.
-  const double STEP_SIZE = 5e-2;
+  const double STEP_SIZE = 5e-5;
+
+  //number of cells in the LSTM (hidden layers in standard terms)
+  //NOTE: you may play with this variable in order to further optimize the model
```
rcurtin: We could also add a note here saying that as more cells are added, accuracy is likely to go up, but training time may take longer.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +61,114 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
 {
-  /* HYPERPARAMETERS */
+  //TODO: change this to your base data path!
+  const string basedatapath = "D:/sample-ml-app";
+
+  //TODO: if true the model will be trained; if false the saved model will be read and used for prediction
+  //NOTE: training the model may take a long time, therefore once it is trained you can set this to false
+  //    and use the model for prediction.
+  //NOTE: there is no error checking in this example to see if the trained model exists!
+  const bool bTrain = true;
```
rcurtin: The previous setup actually allowed loading a model and then training it further... I don't have any issue with keeping the example simple by omitting that functionality and making it either/or, like you've done here, but I want to make sure that's what you intended to do first---so, do you think it's useful to keep that older support, or is it fine as-is?  (I have no real opinion.)

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -84,105 +179,118 @@ int main()
...
+    cout << "Training ..." << endl;
 
-  cout << "Training ..." << endl;
-  // Cycles for monitoring the process of a solution.
-  for (int i = 0; i < EPOCH; i++)
-  {
-    // Train neural network. If this is the first iteration, weights are
-    // random, using current values as starting point otherwise.
-    model.Train(trainX, trainY, optimizer);
+    // Run EPOCH number of cycles for optimizing the solution.
```
rcurtin: Using the new ensmallen callbacks, we can actually clean this up so that the loss is printed after every epoch, and then we can just call `model.Train()` once only, with `optimizer.MaxIterations()` set to `EPOCH * trainX.n_cols`.  But there's no need to do that here---we can open an issue for it.  (Unless you want to give it a shot! :))

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: Thank you!

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: OK!

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: Good idea! I will add this extra possibility. The problem with the previous setup was that it was always training and you could not just load and use the model for prediction.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: I have not followed that development. I prefer to keep the code as it is now for the moment.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: 1. OK for the paths! I will add this.

2. Saving and showing the results
> I think it's not too hard to simplify. :) Here's a proposal (please double-check if it works---I haven't run it):

I see what you try to do but this will not work because you have 2 more rows here than in the original file. You would need to inverse transform the input and the prediction separately. You also need to remove the last column because that record was not used for training (no next day for prediction) otherwise you will have the wrong prediction.
I will modify the code according to your suggestion and in a way that it is correct.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: Sounds good---I'll open an issue for it once we merge this code. :+1:

zsogitbe:
I have added the improved LSTM/TimeSeries-Univariate tutorial to the pull request.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +62,111 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
   // Testing data is taken from the dataset in this ratio.
   const double RATIO = 0.1;
 
-  // Number of cycles.
-  const int EPOCH = 256;
+  // Number of optimization cycles.
+  const int EPOCH = 500;
 
-  // Number of iteration per epoch.
+  // Number of iterations per cycle.
```
rcurtin: Why not keep this as `epoch`?

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -84,105 +177,127 @@ int main()
...
   data::MinMaxScaler scale;
   scale.Fit(dataset);
   scale.Transform(dataset, dataset);
 
-
+  //we have 5 input data columns and 2 output columns (target)
   size_t inputSize = 5, outputSize = 2;
 
+  //We need to represent the input data for RNN in arma::cube (3D matrix)! The 
+  //3rd dimension is the rho number of past data records the RNN uses for
```
rcurtin:
```suggestion
  //3rd dimension is rho, the number of past data records the RNN uses for 
```

Let me know if I missed something; this seems a little clearer to me. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -84,105 +177,127 @@ int main()
...
-  cout << "Finished" << endl;
-  cout << "Saving Model" << endl;
-  if (saveModel)
-  {
-    data::Save("saved_models/LSTMMulti.bin", "LSTMMulti", model);
-    std::cout << "Model saved in saved_models/." << std::endl;
+    cout << "Finished training." << endl;
+    cout << "Saving Model" << endl;
+    data::Save(modelFile, "LSTMMulti", model);
+    std::cout << "Model saved in " << modelFile << std::endl;
```
rcurtin: Up to you---you can remove the `std::` from wherever it appears throughout the file, or I can do that during merge. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -84,105 +177,127 @@ int main()
...
   }
+
+  //NOTE: the below is added in order to show how in a real application the 
+  //model would be saved, loaded and then used for prediction. Please note that 
+  //we do not have the last data point in testX because we did not use it for 
+  //the training, therefore the prediction result will be for the day before!
```
rcurtin: Do you mean here that the prediction results don't consider the last day in the dataset?

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +62,111 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
+  //The prediction results are the (high, low) for the next day and comming from 
+  //the last slice from the prediction.
+  mat temp = predictions.slice(predictions.n_slices - 1);
+
+  //NOTE: We add 3 extra rows here in order to recreate the input data 
+  //structure used to transform the data. This is needed in order to be able 
+  //to use the right scaling parameters for the specific column stock 
+  //high, low).
+  temp.insert_rows(0, 3, 0);
+  scale.InverseTransform(temp, temp);
```
rcurtin: Let me know if I am overlooking something here.  It seems to me that we are first calling `scale.InverseTransform` with `flatDataAndPreds` to convert back the first 3 rows, and then calling again with `temp` to convert back the predictions.  Couldn't we do this all at once?  Like this:

```
mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
flatDataAndPreds.insert_rows(predictions.slice(predictions.n_slices - 1));
scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
```

Let me know if I overlooked something here.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -84,105 +177,127 @@ int main()
...
+  modelP.Predict(testX, predOutP);
+  // Calculating mse on prediction.
+  double testMSEP = MSE(predOutP, testY);
+  cout << "Mean Squared Error on Prediction data points:= " << testMSEP << endl;
+
+  //save the output predictions and show the results
+  saveAndResults(predFile, predOutP, scale, testX);
+
+  cout << "Ready!" << std::endl;
+  getchar();
```
rcurtin: I know that this is necessary on Windows, because typically this will be run in a command prompt window which would close the moment the program exits, but this won't be the case on Linux or OS X.  We could do something like this if you don't think it makes the code too messy:

```
  // Wait for the user to end the program, but only on Windows, so that if this is run in
  // a command prompt window that the window won't close prematurely.
  #ifdef _WIN32
    cout << "Ready!" << endl;
    getchar();
  #endif
```

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: The naming convention was not consistent and therefore I have chosen to use cycle here. On other places 'cycle' is used.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: OK! This was actually from the original text.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: OK! I have removed it.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: Yes because we use the test data for prediction and the last day was in the test data removed because there is no next day to test/predict.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: What you have overlooked is that the structure of flatDataAndPreds is already OK, we can inverse transform it. But the structure of temp is not yet OK and we must still add 3 rows to it before we can inverse transform it. We can only use the inverse transform if the number of rows equals to the number of rows in the data set used to create the MinMaxScaler. This is the reasons why we need to do this separately.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: OK! I see what you mean. I have commented out that part of the code like this:

  //Use this on Windows in order to keep the console window open
  //cout << "Ready!" << endl;
  //getchar();

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: No worries, good to fix it anyway.  Thanks for the fix. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: I'd prefer to go with `epoch` since that matches the neural network literature a lot better.  If you prefer, I'm happy to make that change during merge.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: Ok, perfect, thank you for the clarification.  I might reword it slightly during merge if you don't mind. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -84,105 +177,127 @@ int main()
...
+  //representation it is the first column.
+  //The first column in the CSV is the date which is not required, therefore 
+  //removing it also (first row in in arma::mat).
   dataset = dataset.submat(1, 1, dataset.n_rows - 1, dataset.n_cols - 1);
 
-  // Scale data for increased numerical stability.
+  // Scale all data into the range (0, 1) for increased numerical stability.
   data::MinMaxScaler scale;
   scale.Fit(dataset);
   scale.Transform(dataset, dataset);
```
rcurtin: Now that I am thinking about it, technically we actually should only do the scaling on the training set. :)  But maybe it is better to just open an issue for that and let someone else take care of it.  It would be good to get this merged soon so I don't want to hold it up with extraneous requests.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: I see.  I dug more closely through the code and I think I may have found another issue (not yours, I guess it was there before) and left a comment there.  Let me know what you think.  Honestly I think once we can resolve this little issue everything is good to merge here.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -43,37 +62,111 @@ double MSE(arma::cube& pred, arma::cube& Y)
...
+  typename LabelType = arma::cube>
 void CreateTimeSeriesData(InputDataType dataset, DataType& X, LabelType& y, size_t rho)
 {
-  for(size_t i = 0;i < dataset.n_cols - rho - 1 ;i++)
+  for (size_t i = 0; i < dataset.n_cols - rho; i++)
   {
-    X.subcube(span(), span(i), span()) = dataset.submat(span(), span(i, i+rho-1));
-    y.subcube(span(), span(i), span()) = dataset.submat(span(3,4), span(i, i+rho-1));
+    X.subcube(span(), span(i), span()) = dataset.submat(span(), span(i, i + rho - 1));
+    y.subcube(span(), span(i), span()) = dataset.submat(span(3, 4), span(i + 1, i + rho));
```
rcurtin: Ahh, now that I've dug through, this bit confuses me.  I would have assumed that for the training set, we would take everything *except* the columns that we are predicting.  But here the use of `span()` on the right hand side of

```
    X.subcube(span(), span(i), span()) = dataset.submat(span(), span(i, i + rho - 1));
```

means we take *all* rows as the training set---including the thing we are trying to predict in `y`!

So, I think maybe this should be:

```
X.subcube(span(), span(i), span()) = dataset.submat(span(0, 2), span(i, i + rho - 1));
```

This would mean that we are using the first three rows as input to predict the last two, which is what I would expect.

Right now, I believe that we are actually passing the outputs we desire to predict in as the input, meaning that the model will do extremely well since we are letting it cheat and see the answer too. :)

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: Ah, even better idea, that keeps the code easier to understand. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: Be careful here because the model learns the scaled data, therefore you need to use scaled test data!

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: No we are not cheating here because the model learns the next step and it is normal to use all rows (all available data at the current step). What the model learns is shifted one record:

`y.subcube(span(), span(i), span()) = dataset.submat(span(3, 4), span(i + 1, i + rho));`

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
zsogitbe: That is not an issue. I explained it below. By the way, you may remove my name from the code if you prefer! What I find important is that the completely wrong tutorial is corrected and may be very useful to many!

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: Right, exactly, just the `Fit()` call should be on the training subset.  Anyway, I'll open an issue for it after this gets merged. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: You're absolutely right.  Thank you for pointing that out!  Not sure why I missed that last night.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: You're right, it's not an issue.  Thanks for clarifying that.  But the original question I had in this thread still remains:

I believe we can just do one call to `InverseTransform()`; in the end, all we need for this method is the unscaled *predictions*, not the unscaled features.  So, there are a couple of ways we can do that.  We can do what you're already doing here, but it's a bit redundant because it calls `InverseTransform()` twice, and none of the results from the first call to `InverseTransform()` are actually used.

We could also simply omit the first call to `InverseTransform()` and do this:

```
 mat temp = predictions.slice(predictions.n_slices - 1);

  //NOTE: We add 3 extra rows here in order to recreate the input data 
  //structure used to transform the data. This is needed in order to be able 
  //to use the right scaling parameters for the specific column stock 
  //high, low).
  temp.insert_rows(0, 3, 0);
  scale.InverseTransform(temp, temp);
```

It would be possible to just use `temp` now in the printing code on lines 126 to 128.  The third idea, which is a corrected version of what I suggested above, would be to take `testX`, replace the last two columns with the output predictions, then call `InverseTransform()`:

```
mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
flatDataAndPreds.rows(testX.n_rows - 2, testX.n_rows - 1) = predictions.slice(predictions.n_slices - 1);
scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
```

(We'd still need to drop the last column in all of these cases.)

I think this change would simplify the code, and at this point I'm pretty sure either of the alternatives above are right (and I'm fine with either of them), but I want to double check with you that this wouldn't subtly introduce an error.  So, let me know what you think.

zsogitbe: Thank you for explaining this! It is very insightful! I however do not have more time to invest in this example and this issue is not very important.

rcurtin: Ok, no problem.  I can make the change during merge, I just want to check that you see no issues if I do that.

zsogitbe: I do not approve this change because it is not tested. You better leave it like it is.

rcurtin: Ok, no problem, I'll open another issue after this is merged.

zoq:
Already glanced over the changes, will take a closer look tomorrow.

zoq:
>  Those will be nice issues for anyone looking to contribute some simple patches.

I could also do the changes, and we use this as an example, for other models.

rcurtin:
@zsogitbe thank you, I merged things in and fixed some style issues in 84788bc.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: I opened #41 for this.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: I opened #42 for this.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: I opened #43 for this.

rcurtin:
> > Those will be nice issues for anyone looking to contribute some simple patches.
> 
> I could also do the changes, and we use this as an example, for other models.

I opened a few issues, so if you want to do them, feel free, otherwise maybe someone else will get to them. :+1:

PR #39, opened 2020-01-21
Modernization of the cmake in [Model]

knakul853:
I tried my hand at the "modernizing" model's Cmake. This is the first time I've worked with a build system (or rather a build system generator) so I might have made silly mistakes or misunderstood some things, so bear with me :).

I mostly just reconfigured the code to work at a target level instead of a directory level.

- follow the latest CMake version(3.16) syntax.
- add target-properties  bassed  syntax to CMake
- add multiline comments
- add more boost version 
- more CMake new feature [cmake](https://cliutils.gitlab.io/modern-cmake/)


rcurtin:
Can you put descriptions with your PRs please?  For instance, why would we want to merge this change?  What is it even changing?  Why are the changes necessary?  Thanks. :+1:

CMakeLists.txt
```diff
@@ -1,10 +1,18 @@
-cmake_minimum_required(VERSION 2.6)
-project(models C CXX)
+cmake_minimum_required(VERSION 3.10)
+
+# set the project name and version
```
zoq: Looks like the idea here is to use some of the new CMake features, I think we can do a lot more similar to: https://github.com/mlpack/ensmallen/pull/152/files

knakul853: @zoq  yeah I mistakenly open the PR I was just playing with the codebase. Now i would complete this.

.travis.yml
```diff
@@ -5,7 +5,7 @@ language: cpp
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq libopenblas-dev liblapack-dev g++ libboost-all-dev
-  - curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-6.500.5.tar.gz | tar xvz && cd armadillo*
+  - curl http://sourceforge.net/projects/arma/files/armadillo-9.800.2.tar.xz | tar xvz && cd armadillo*
```
zoq: You could use:

https://github.com/mlpack/ensmallen/blob/51614c216e43b235b97ff9059f571c8f02e3de37/.travis.yml#L16

as done in the ensmallen repo.

.travis.yml
```diff
@@ -2,12 +2,20 @@ sudo: required
 dist: trusty
```
zoq: Maybe if we run on Xenial it will help.

knakul853:
Hey, I think the falling of build is not related to the CMake.what do you think ?

.travis.yml
knakul853: @zoq  tons of thanks : )

knakul853:
@zoq @rcurtin  please add some comments : )

knakul853:
I will be rewriting the CMake as @rcurtin  suggested.

PR #48, opened 2020-02-07
[WIP] Make Appveyor Use Visual Studio

birm:
So, it's kind of pointless to just use appveyor as a second linux CI system. So, this PR intends to (try to) continue on the work of #10 to get a VS CI test working.
I suspect I'll be waiting for builds to queue/fail a lot.


birm:
Okay, so now I've duplicated commits but kept the blame ok? This is insane. I'll need to fix this commit mess once I've fixed VS.

zoq:
Not sure if it makes things easier, but I think you can ssh into the build node.

birm:
So, it looks like I'm occupying valuable appveyor time from the mlpack organization. So, what I've done is create my own account, so I can test without interference. I'll re-open this once I get this working.

birm:
I'm keeping this closed. There's not enough time to build both mlpack and models from source, and I can't get the mlpack artifacts to work right (I think that'd be the way to do it if anyone else wants to take over).

rcurtin:
Hmmm, the only part of mlpack that needs to be built is just mlpack.dll, so maybe we can configure CMake to avoid the time-consuming stuff: `-DBUILD_CLI_EXECUTABLES=OFF -DBUILD_TESTS=OFF -DBUILD_PYTHON_BINDINGS=OFF`. Maybe that helps?

PR #49, opened 2020-02-08
Rewriting cmake 

knakul853:
I tried my hand at the "modernizing" model's Cmake. This is the first time I've worked with a build system (or rather a build system generator) so I might have made silly mistakes or misunderstood some things, so bear with me :).

I mostly just reconfigured the code to work at a target level instead of a directory level.

- follow the latest CMake version(3.16) syntax.
- add target-properties  bassed  syntax to CMake
- add multiline comments
- add more boost version 
- more CMake new feature [cmake](https://cliutils.gitlab.io/modern-cmake/)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "files.associations": {
```
zoq: Hm, this looks like a editor specific config file to me.

CMakeLists.txt
```diff
@@ -183,26 +315,47 @@ if (NOT ENSMALLEN_FOUND)
...
-      unit_test_framework
-      serialization
-    REQUIRED
-)
+set( BOOST_VERSION 1.72.0 )
+# Settings for finding correct Boost libraries.
+set( Boost_USE_STATIC_LIBS FALSE )
+set( Boost_USE_MULTITHREADED TRUE )
+set( Boost_USE_STATIC_RUNTIME FALSE )
+set( Boost_COMPILER "-gcc8" )
```
zoq: Is this needed, there are various compiler out there.

knakul853:
@zoq @birm  how do I make sure to download boost the latest version(1.72.0). After adding the latest PPA it seems to not working.

birm:
You shouldn't need the latest boost. Which version do you have? How did you install it?

knakul853:
@birm on local machine I have installed boost 1.72.0 and made some change in cmake according to latest boost.i want it to be added on the .Travis.yaml and tested.

birm:
@knakul853 is there something about boost 1.72 that you need in particular? We should be able to use boost as old as 1.49.

knakul853:
> @knakul853 is there something about boost 1.72 that you need in particular? We should be able to use boost as old as 1.49.

.travis.yml
```diff
@@ -2,6 +2,10 @@ sudo: required
 dist: xenial
 language: cpp
 
+env:
+  - ARMADILLO=latest
+  - ARMADILLO=minimum
+
```
birm: It looks like we're using armadillo 8.400 no matter what; are we using these environment variables elsewhere?

CMakeLists.txt
```diff
@@ -183,26 +315,46 @@ if (NOT ENSMALLEN_FOUND)
...
-set(Boost_ADDITIONAL_VERSIONS
-  "1.49.0" "1.50.0" "1.51.0" "1.52.0" "1.53.0" "1.54.0" "1.55.0")
-find_package(Boost 1.49
-    COMPONENTS
-      program_options
-      unit_test_framework
-      serialization
-    REQUIRED
-)
+set( BOOST_VERSION 1.58.0 )
```
birm: Does boost 1.49.0 no longer work?

CMakeLists.txt
```diff
@@ -250,3 +396,5 @@ set(DIRS
 foreach(dir ${DIRS})
   add_subdirectory(${dir})
 endforeach()
+
+enable_testing()
```
birm: might we want to put this in a conditional with some flag such as "BUILD_TESTING", or do we always want to build tests?

CMake/FindArmadillo.cmake
```diff
@@ -365,5 +365,12 @@ endif ()
 mark_as_advanced(
   ARMADILLO_INCLUDE_DIR
   ARMADILLO_LIBRARY)
+  
+if (ARMADILLO_FOUND AND NOT TARGET Armadillo:Armadillo)
+  add_library(Armadillo::Armadillo INTERFACE IMPORTED)
+  set_target_properties(Armadillo::Armadillo PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${ARMADILLO_INCLUDE_DIR}"
+      INTERFACE_LINK_DIRECTORIES "${SUPPORT_INCLUDE_DIRS}" INTERFACE_LINK_LIBRARIES "${ARMADILLO_LIBRARIES}")
+endif()    
+
```
birm: As an aside, I'm curious on where you found guidance on modern use of the FindX.cmake scripts, since I haven't seen much of that myself.

birm:
Great work! What platforms/operating systems has it been tested with? Do you need any help testing it?

CMake/FindArmadillo.cmake
```diff
@@ -365,5 +365,12 @@ endif ()
 mark_as_advanced(
   ARMADILLO_INCLUDE_DIR
   ARMADILLO_LIBRARY)
+  
+if (ARMADILLO_FOUND AND NOT TARGET Armadillo:Armadillo)
```
rcurtin: Should this be `Armadillo::Armadillo`?

CMakeLists.txt
```diff
@@ -1,72 +1,36 @@
...
-    message(FATAL_ERROR "No C++11 compiler available!")
-  endif()
-  enable_cxx11()
-endif ()
-
-# Include modules in the CMake directory.
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/CMake")
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
+# specify the minimum version. 
+cmake_minimum_required(VERSION 3.1...3.16)
```
rcurtin: Why is there a version range for the minimum required version?

CMakeLists.txt
```diff
@@ -1,72 +1,36 @@
...
 
-# Set the CFLAGS and CXXFLAGS.
-# Set the CFLAGS and CXXFLAGS depending on the options the user specified.
-# Only GCC-like compilers support -Wextra, and other compilers give tons of
-# output for -Wall, so only -Wall and -Wextra on GCC.
-if(CMAKE_COMPILER_IS_GNUCC OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0 -Wall -Wextra -ftemplate-depth=1000")
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
+# set cmake policy.
```
rcurtin: Can you use a more descriptive comment here?

CMakeLists.txt
```diff
@@ -1,72 +1,36 @@
...
-  IF (APPLE)
-    # detect OS X version. Use '/usr/bin/sw_vers -productVersion' to
-    # extract V from '10.V.x'.)
-    EXEC_PROGRAM(/usr/bin/sw_vers ARGS
-        -productVersion OUTPUT_VARIABLE MACOSX_VERSION_RAW)
-    string(REGEX REPLACE
-        "10\\.([0-9]+).*" "\\1"
-        MACOSX_VERSION
-        "${MACOSX_VERSION_RAW}")
+# set the standards
```
rcurtin: This comment could be more descriptive too.  It's pretty clear from the code itself that it's setting the standards, but an effective comment would show *why* we are doing this.  e.g.,

```
# We require C++11 because that's what mlpack requires.
```

or something like that.  This applies to most of the comments in this file; please go through and rework them---it'll make this a lot easier to maintain in the long run. :+1:

CMakeLists.txt
```diff
@@ -86,9 +50,9 @@ if(CMAKE_SIZEOF_VOID_P EQUAL 8)
       # We are on a 64-bit system.  Does Armadillo have ARMA_64BIT_WORD enabled?
       file(READ "${ARMADILLO_INCLUDE_DIR}/armadillo_bits/config.hpp" ARMA_CONFIG)
       string(REGEX MATCH
-          "[\r\n][ ]*#define ARMA_64BIT_WORD"
-          ARMA_HAS_64BIT_WORD_PRE
-          "${ARMA_CONFIG}")
+                   "[\r\n][ ]*#define ARMA_64BIT_WORD"
+             ARMA_HAS_64BIT_WORD_PRE
+                    "${ARMA_CONFIG}")
```
rcurtin: Why the extra indentation?

CMakeLists.txt
```diff
@@ -110,6 +74,174 @@ else()
...
+else ()
+  set(CMAKE_INSTALL_BINDIR ${CMAKE_INSTALL_PREFIX}/bin)
+  set(CMAKE_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX}/lib)
+  set(CMAKE_INSTALL_MANDIR ${CMAKE_INSTALL_PREFIX}/man)
+  set(CMAKE_INSTALL_DOCDIR ${CMAKE_INSTALL_PREFIX}/share/doc/mlpack)
+  set(CMAKE_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_PREFIX}/include)
+endif ()
+
+# This is as of yet unused.
+option(PGO "Use profile-guided optimization if not a debug build" ON)
```
rcurtin: This is unused, so honestly we could just remove it.

CMakeLists.txt
```diff
@@ -183,26 +315,46 @@ if (NOT ENSMALLEN_FOUND)
   endif ()
 endif ()
 
-# Unfortunately this configuration variable is necessary and will need to be
 # updated as time goes on and new versions are released.
```
rcurtin: Leaving just this line as a comment doesn't make much sense.

CMakeLists.txt
rcurtin: Right, we should stick with 1.49 since that's the minimum required version for mlpack.

CMakeLists.txt
```diff
@@ -183,26 +315,46 @@ if (NOT ENSMALLEN_FOUND)
...
-      unit_test_framework
-      serialization
-    REQUIRED
-)
+set( BOOST_VERSION 1.58.0 )
+# Settings for finding correct Boost libraries.
+set( Boost_USE_STATIC_LIBS FALSE )
+set( Boost_USE_MULTITHREADED TRUE )
+set( Boost_USE_STATIC_RUNTIME FALSE )
+set( Boost_ADDITIONAL_VERSIONS "${BOOST_VERSION}" )
```
rcurtin: These settings likely aren't portable, and the user can specify them from the command line, so we should omit them.

CMakeLists.txt
```diff
@@ -183,26 +315,46 @@ if (NOT ENSMALLEN_FOUND)
...
+
+find_package(Boost ${BOOST_VERSION}
+      MODULE
+      REQUIRED COMPONENTS program_options
+                          unit_test_framework
+                          serialization
+                          graph)
+# Make found targets globally available.
+if ( Boost_FOUND )
```
rcurtin: A style issue---please do `if (Boost_FOUND)` instead of `if ( Boost_FOUND )`.  This type of comment applies throughout the code. :+1:

CMakeLists.txt
rcurtin: I wasn't aware there were any tests in this repository---I thought it was all just programs.  So I'm not sure this line is needed.

LSTM/TimeSeries-Multivariate/CMakeLists.txt
```diff
@@ -1,11 +1,9 @@
-cmake_minimum_required(VERSION 2.6)
-project(lstm-time-series)
+cmake_minimum_required(VERSION 3.1...3.16)
 
+project(lstm-time-series)
```
rcurtin: Should this really be a separate project?

CMake/FindArmadillo.cmake
knakul853: Hey sorry for the late reply.There was exams.
For reference I used this resources https://cliutils.gitlab.io/modern-cmake/

CMakeLists.txt
knakul853: Ok thanks

.travis.yml
knakul853: No, let me delete this.

CMakeLists.txt
knakul853: yeah, for now, I don't think we need also,.,will remove

CMakeLists.txt
knakul853: If the CMake version is less than 3.12, then if block will be true, and the policy will be set to the current CMake version. If CMake is 3.12 or higher, the if block will be false, but the new syntax in cmake_minimum_required will be respected and this will continue to work properly.

knakul853: In CMake 3.12, this will support a range, such as VERSION 3.1...3.12; this means you support as low as 3.1 but have also tested it with the new policy settings up to 3.12. This is much nicer on users that need the better settings, and due to a trick in the syntax, it's backward compatible with older versions of CMake (though actually running CMake 3.2-3.11 will only set the 3.1 version of the policies in this example). New versions of policies tend to be most important for macOS and Windows users, who also usually have a very recent version of CMake.

LSTM/TimeSeries-Multivariate/CMakeLists.txt
knakul853: ofcourse not..XD

knakul853:
Sorry for the late response. After exams, I am good to go.

CMakeLists.txt
knakul853: added

CMakeLists.txt
knakul853: added

CMakeLists.txt
knakul853: fixed

favre49:
I think this would apply to the `models` repository, since this repository will be undergoing a big overhaul. Sorry for the inconvenience, but could you migrate it there? (Granted `models` structure will probably change a lot too)

birm:
Since this repository no longer uses cmake, closing.

PR #55, opened 2020-02-20
Add callback to monitor the training process

shrit:


Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -132,41 +130,47 @@ int main()
...
-  // should be sent to Kaggle website).
+  model.Train(trainX, 
+              trainY,
+              optimizer,
+              ens::PrintLoss(),
+              ens::ProgressBar(),
+              ens::EarlyStopAtMinLoss(),
+              ens::StoreBestCoordinates<arma::mat>());
+
+  // don't reset optimizer's parameters between cycles.
```
zoq: Picky style comment, the comment should start with a capital letter.

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -132,41 +130,47 @@ int main()
     // Adam update policy.
     AdamUpdate(1e-8, 0.9, 0.999));
 
-  // Cycles for monitoring the process of a solution.
-  for (int i = 1; i <= CYCLES; i++)
-  {
+  // Relay on Early Stopping as stopping criterion
```
zoq: Missing stop at the end.

shrit: I am sorry, but I did not get what do you mean here.

zoq: There is a missing `.` at the end; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments for more information.

shrit:
@zoq  Would you think that it is better to have some program options in this example, just to allow the users to pass the path of the model dynamically instead of compiling the model each time. It seems that model zoo requires boost program options as a dependency. If it is the case, I can use it to so that, or I can do it old fashin (argc, argv) function?

zoq:
I think that is a great idea, I guess, we can also use mlpack's CLI functionality for the same. I guess that would bring us one step closer to expose those models to other languages as well (bindings for python, julia, etc.).

shrit:
Great, I will push as soon as possible with mlpack's CLI

kartikdutt18:
@shrit, will you be pursing the same for LSTM or do you mind if I take that up?

shrit:
I am going to finish only the Kaggle models. You can start working on LSTM

kartikdutt18:
Great, I hope to open a PR soon.

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -79,8 +144,9 @@ int main()
       train.n_cols - 1) / 255.0;
   const mat validX = valid.submat(1, 0, valid.n_rows - 1,
       valid.n_cols - 1) / 255.0;
-  
-  const int ITERATIONS_PER_CYCLE = trainX.n_cols;
+
+  // Allow infinite number of iterations utill we stopped by EarlyStopAtMinLoss.
```
gaurav-singh1998: Minor spelling issue, utill -> until.

shrit:
@zoq I think there is something missing with the actual CMake. Since I have added the mlpack:CLI It was not possible to link with boost program option, (my version is 1.71.0). CMake was able to find the boost headers, but not the implementations.
In order to compile on my machine, I had to use `set(program_options "/usr/lib64/libboost_program_options.so")` and link with `program_options`
Also travis is complaining about similar issue.

rcurtin:
Ahhh, sorry to be a little late to the party here and reply after a lot of the code has already been done.  At least my understanding of the `models` repo is that it contained self-contained examples that people could use and adapt for their tasks.  Using the mlpack CLI framework, in my opinion, makes it a good bit harder to actually do that---instead of being able to just copy-paste the code in the example that concerns the model training, now they have to understand what all the CLI stuff is.  Also, it would be a lot harder to compile the program by hand, since the `BINDING_TYPE` macro would need to be defined correctly.

I guess, what I mean is, the use of the CLI class here does mean that we can nicely compile into a program that can do digit recognition with various parameters.  But I think that more people are using this as a simple example, moreso than actually doing digit recognition with it.

I think that it would be better to have a "generic" binding in the main mlpack repository that can work with neural networks of arbitrary shape and size (defined in a text file or something).  There's an issue open for that: mlpack/mlpack#1254.  Then, the models repository can primarily be a set of minimal-code examples that people can adapt into their own work.

shrit:
@rcurtin I understand.
No problem, I can remove mlpack::CLI. However, it is a little bit strange to put the path to the training dataset directly in the code, the user will have to re-compile each time for a new path, no forgetting that the compilation time is considerable.
We can at least use the old C++ style `(argc, argv)` for these paths. Or we can leave it just like that. Please tell me what you find the best option for the user.
For the above issue https://github.com/mlpack/mlpack/issues/1254, creating something similar might be possible, but there are some points that need be detailed:
1) The issue seems to be related to https://github.com/mlpack/mlpack/pull/1837. Does @sreenikSS still working on it?
2) @kunakl07 Are you working on it?
3) @sreenikSS stated that the issue https://github.com/mlpack/mlpack/pull/1837 can go to stale since it would be part of _model converter_ repository, does this repository exist, if yes, is it public, is there any issue opened detailing this converter.

Finally, if the converter is ready, we will have only to add mlpack::CLI and write the `mlpack_ann` main example which should be easier since the hard part (the parser) have been accomplished

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -23,164 +24,236 @@
...
+            << "\t -h, --help \t Show this help message." << std::endl
+            << "\t -t, --training_dataset \t Full path to the file containing "
+               "the training set."
+            << std::endl
+            << "\t -l, --testing_datatest, \t Full path to the file containing "
+               "the test set."
+            << std::endl;
+}
+
+int main(int argc, char* argv[])
```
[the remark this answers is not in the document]
shrit: I agree,  let us wait for the discussion in https://github.com/mlpack/models/issues/61 to see what to do with `argc, argv`
You do not have to worry about it, It was not much effort to do remove all the CLI code.

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -23,164 +24,236 @@
...
-
-    cout << i << " - accuracy: train = "<< trainAccuracy << "%," <<
-      " valid = "<< validAccuracy << "%" <<  endl;
-  }
+      // Step size of the optimizer.
+      STEP_SIZE,
+      // Batch size. Number of data points that are used in each iteration.
+      BATCH_SIZE,
+      // Max number of iterations
+      ITERATIONS_PER_CYCLE,
```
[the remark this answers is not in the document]
shrit: Agreed

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -23,164 +24,236 @@
 #include <Kaggle/kaggle_utils.hpp>
 
 #include <ensmallen.hpp>
+#include <ensmallen_bits/callbacks/callbacks.hpp>
 
-using namespace mlpack;
```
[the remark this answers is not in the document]
shrit: In fact, I do not know mlpack convention for namespaces. If you would like I will keep all of them. However, I would not use any `using namespace` in his case since there are several libraries, Actually, when I started copying the code as a simple user, I had a problem knowing to which library each function belongs, especially there are the Kaggles functions that do not belong to any namespace.
For example, I would only keep namespace if we were using only mlpack. What do you think?

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -23,164 +24,236 @@
...
+
+  // Train neural network. If this is the first iteration, weights are
+  // random, using current values as starting point otherwise.
+  model.Train(trainX,
+              trainY,
+              optimizer,
+              ens::PrintLoss(),
+              ens::ProgressBar(),
+              ens::EarlyStopAtMinLoss(),
+              ens::StoreBestCoordinates<arma::mat>());
```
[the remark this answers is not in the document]
shrit: Agreed

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -23,164 +24,236 @@
...
+
+  mat predOut;
+  // Getting predictions on training data points.
+  model.Predict(trainX, predOut);
+  // Calculating accuracy on training data points.
+  Row<size_t> predLabels = getLabels(predOut);
+  double trainaccuracy = accuracy(predLabels, trainY);
```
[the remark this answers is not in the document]
shrit: I agree. All the helper functions defined here https://github.com/mlpack/models/blob/master/Kaggle/kaggle_utils.hpp can be replaced, especially since we have callbacks know. At the same time, the idea of an accuracy function as a callback is not a bad one for all classification problems. What do you think?

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -23,164 +24,236 @@
 #include <Kaggle/kaggle_utils.hpp>
 
 #include <ensmallen.hpp>
+#include <ensmallen_bits/callbacks/callbacks.hpp>
 
-using namespace mlpack;
 using namespace mlpack::ann;
 
 using namespace arma;
-using namespace std;
```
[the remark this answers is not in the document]
shrit: Same as before, even I would declare a namespace for these helpers https://github.com/mlpack/models/blob/master/Kaggle/kaggle_utils.hpp if we decide to keep them. In this file it defines a `save` function, which can be ambiguous, since there are `mlpack::data::save` and `arma::save`

shrit:
@rcurtin I will remove `argc, argv` from this code. In order to be compatible with examples.
I left some comments on some questions considering the namespaces. 
I will apply callbacks on `vae`. Also, I will re-apply the same pull request on the new model repository since the title only refer to callbacks. 
We will look into details later in the model repository considering all things related to `mlpack::CLI`

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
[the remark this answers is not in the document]
rcurtin: Should we make a change here then?  We can just drop `StoreBestCoordinates` if you like, that would be fine. :+1:

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
```diff
@@ -41,146 +41,139 @@ int main()
...
+  // Getting predictions on validating data points.
+  model.Predict(validX, predOut);
+  // Calculating accuracy on validating data points.
+  predLabels = getLabels(predOut);
+  double validAccuracy = arma::accu(predLabels == validY) / trainY.n_elem;
+
+  std::cout << "Accuracy: train = " << trainAccuracy << "%,"
+            << " Valid = " << validAccuracy << "%" << endl;
+
+  mlpack::data::Save("model.txt", "model", model, false);
```
rcurtin: You could also consider `model.bin` for a more compressed representation.  I suspect the text format won't be very readable. :smile: (It may actually have precision issues representing floating-point numbers, too.)

vae/vae_mnist/src/vae.cpp
```diff
@@ -56,9 +56,7 @@ int main()
   // The step size of the optimizer.
   constexpr double stepSize = 0.001;
   // The number of interations per cycle.
-  constexpr int iterPerCycle = 56000;
-  // Number of cycles.
-  constexpr int cycles = 100;
+  constexpr int maxIteration = 0;
```
rcurtin: It might make sense to change the comment now to point out that this is the total maximum number of iterations, and that `0` represents no limit.  (Also, is that what we want for this example?)

vae/vae_mnist/src/vae.cpp
```diff
@@ -162,19 +160,17 @@ int main()
...
-    // Don't reset optimizer's parameters between cycles.
-    optimizer.ResetPolicy() = false;
-
-    std::cout << "Loss after cycle " << i << " -> " <<
+  vaeModel.Train(train,
+                 train,
+                 optimizer,
+                 ens::PrintLoss(),
+                 ens::ProgressBar(),
+                 ens::EarlyStopAtMinLoss());
```
rcurtin: I suppose then that this runs until `EarlyStopAtMinLoss` terminates the optimization?  Probably it would be good to add a comment explaining that too. :+1:

shrit: Agreed

vae/vae_mnist/src/vae.cpp
shrit: Agreed

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
shrit: It is good to know that `.bin` might be better than `.txt`.

Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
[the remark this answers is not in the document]
shrit: @rcurtin Tell me what do you think about `namespaces`. I would remove all of them if it is fine. What do you think?

shrit:
@rcurtin Yes I going to open a new pull request for models with all the `CLI` binding for the MNIST models

shrit:
@rcurtin I think you can merge this one now. 
If there is any problem we can handle it in a new pull request. 
The objective of this pull request is accomplished.

mnist_simple/mnist_simple.cpp
```diff
@@ -41,146 +39,139 @@ int main()
...
+            << " Valid = " << validAccuracy << "%" << endl;
+
+  mlpack::data::Save("model.bin", "model", model, false);
 
   // Loading test dataset (the one whose predicted labels
-  // should be sent to Kaggle website).
-  // As before, it's necessary to get rid of header.
+  // should be sent to kaggle website).
+  arma::mat testingDataset;
+  mlpack::data::Load("../Kaggle/data/test.csv", testingDataset, true);
```
rcurtin: What should we do about the data?  Now it's in `../data/kaggle_train_test_dataset.zip`.  I guess we could assume that the user has unpacked the zip in that directory?  (Or maybe we provide a quick script to unpack all the data or something?)

shrit: I prefer removing the entire data directory and provide a small script that downloads the necessary data for each example. We do not have to keep data tracked by git. This will increase the size of the git repository for no obvious reason.
In addition, we might use bigger datasets in the future, which can not be added to this repository.

mnist_cnn/mnist_cnn.cpp
```diff
@@ -106,101 +101,99 @@ int main()
...
+  // Get predictions on training data points.
+  model.Predict(trainX, predOut);
+  // Calculate accuracy on training data points.
+  Row<size_t> predLabels = getLabels(predOut);
+  double trainAccuracy = arma::accu(predLabels == trainY) / trainY.n_elem;
+  // Get predictions on validating data points.
+  model.Predict(validX, predOut);
+  // Calculate accuracy on validating data points.
+  predLabels = getLabels(predOut);
+  double validAccuracy = arma::accu(predLabels == validY) / validY.n_elem;
```
kartikdutt18, line 191: Hey @shrit, I think this might not be correct. As They will never will be exactly equal. Do you mind running some test on this. Thanks a lot.

kartikdutt18: I think I am wrong here, they are labels so they should be correct. However when I tried them I got an accuracy of 0 and with existing function I got an accuracy of 96%. Maybe I did something wrong.

shrit: @kartikdutt18 You are right, I missed this file, I need to remove `getLabels` functions as I have removed the `kaggle_utils.hpp` file. I rebased and I did not build locally.
I was working on a set of makefiles for all examples. The problem is that if we use only `Makefiles`instead of CMake, we will lose support for `Visual Studio` what do you think @rcurtin @kartikdutt18

rcurtin: @kartikdutt18 the code here looks right to me, but casting can make a difference here---`size_t` divided by `size_t` will be 0 in this case, but if one of the types is `double` (or if the result type is `double`?) then I think it will work fine.  If the current code gives good accuracy numbers then I think it's fine.  Also, I realized, we should multiply `trainAccuracy` and `validAccuracy` by 100. :)

@shrit good point about Visual Studio.  The real complexity of build systems like CMake is just in getting all the dependencies right and everything.  The basic Makefile we'd provide would just assume that everything was available on the system, so even that is not "bulletproof"---the user will still have to make sure that their system is set up right.  For Visual Studio, I *think* that if the user sets up a new project correctly (like in the first part of https://www.mlpack.org/doc/mlpack-3.2.2/doxygen/sample_ml_app.html), then they can copy-paste this code directly and it should work.

mnist_simple/mnist_simple.cpp
rcurtin: Sounds good, that works for me.  Do you want to open an issue pointing out that we should add a script to download any datasets?

rcurtin:
@shrit @kartikdutt18 let me know when you guys are good with merging this, and I'm happy to hit the button. :)

mnist_cnn/mnist_cnn.cpp
kartikdutt18: Also we will be adding notebooks so users can use that and we can instructions for gcc etc. So maybe that isn't an issue. Not sure though.

kartikdutt18:
> @shrit @kartikdutt18 let me know when you guys are good with merging this, and I'm happy to hit the button. :)

Just one last thing, If an @shrit could add some results in comments about accuracy, I am good to go as well.

shrit:
@kartikdutt18 Agreed, I will try to test this before tomorrow. (I will multipliy by 100 for accuracy).

kartikdutt18:
Great. Thanks a lot.

mnist_simple/mnist_simple.cpp
shrit: I will handle the issue with a pull request directly.
I will try to found the sources for the dataset and download them using wget or curl.

shrit:
@kartikdutt18 This is part of the output related to `mnist_batch_norm` output training process. While it seems that accuracies are `0`, also the model does not learn well. I do not know if these examples have given better results before. 
The `getLables` function seems to work, since the output is between `0->9`. I need to double check everything in this code.
```
Epoch 114/0
0.0137785====================================================================================================] 100% - ETA: 0s - loss: 0.0137465
591/591 [====================================================================================================] 100% - 10s 17ms/step - loss: 0.0137785
Validation loss: 1113.26
Epoch 115/0
0.0162865====================================================================================================] 100% - ETA: 0s - loss: 0.0162486
591/591 [====================================================================================================] 100% - 10s 17ms/step - loss: 0.0162865
Validation loss: 1096.42
Epoch 116/0
0.0172006====================================================================================================] 100% - ETA: 0s - loss: 0.0171606
591/591 [====================================================================================================] 100% - 10s 17ms/step - loss: 0.0172006
Validation loss: 1072.54
Epoch 117/0
0.018189[====================================================================================================] 100% - ETA: 0s - loss: 0.0181468
591/591 [====================================================================================================] 100% - 10s 18ms/step - loss: 0.018189
Validation loss: 1066.69
Epoch 118/0
0.017634[====================================================================================================] 100% - ETA: 0s - loss: 0.017593
591/591 [====================================================================================================] 100% - 10s 18ms/step - loss: 0.017634
Validation loss: 1040.55
Epoch 119/0
0.016979[====================================================================================================] 100% - ETA: 0s - loss: 0.0169395
591/591 [====================================================================================================] 100% - 12s 20ms/step - loss: 0.016979
Validation loss: 1048.35
Epoch 120/0
0.0186793====================================================================================================] 100% - ETA: 0s - loss: 0.0186359
591/591 [====================================================================================================] 100% - 10s 18ms/step - loss: 0.0186793
Validation loss: 1142.12
Accuracy: train = 0%, valid = 0%
Predicting ...
Saving predicted labels to "results.csv"
Neural network model is saved to "model.bin"
Finished

```

kartikdutt18:
Hey @shrit, I have a notebook with mnist example using CNN and it gives decent results. Here is the [link](https://github.com/kartikdutt18/examples/blob/Add-Mnist-Ens-version/mnist_cnn/MNIST_CNN.ipynb). Also the results can also be seen from #74 examples. I will try BatchNorm example as well and get back to you.

shrit:
Yes, but your example uses the old functions from `kaggle_utils.hpp`. I am using also using early stop callback on the validation set.

kartikdutt18:
I think I also used ensmallen callbacks. But yes I used Kaggle_utils functions.

rcurtin:
@shrit maybe you need to cast `trainY.n_elem` to `double`?  I wonder if the division being done is integer division still.  (I don't remember the C++ type promotion rules for mathematical operations.)

shrit:
@rcurtin, Yeap, it was related to double casting, I think we can hit the magic button now.

```
Epoch 18/0
0.0479495====================================================================================================] 100% - ETA: 0s - loss: 0.0478381
591/591 [====================================================================================================] 100% - 18s 30ms/step - loss: 0.0479495
Validation loss: 798.352
Accuracy: train = 99.5026%,      valid = 97.381%
Predicting ...
Saving predicted labels to "results.csv" ...
Neural network model is saved to "model.bin"
Finished

```

rcurtin:
Awesome.  Magic button pressed. :)

PR #56, opened 2020-02-27
Use Ensmallen Callbacks To train LSTM and Fit only on training data.

### LINKED ISSUE #41: Use ensmallen callbacks for LSTM examples
Right now, the optimizations in the LSTM examples manually run a single epoch at a time, in order to print the loss at the end of each epoch.

But now that we have ensmallen callbacks, this isn't necessary anymore, and we can significantly simplify our code.

The task is to go through the LSTM examples and make this change.

This comes from the discussion here: https://github.com/mlpack/models/pull/36#discussion_r362141814

### LINKED ISSUE #42: Simplify `SaveResults()` in LSTM examples
This comes from the discussion here: https://github.com/mlpack/models/pull/36#discussion_r366669401

The goal is to make the change to the code that I suggested, and ensure that the code still works properly (so we should compare the output before and after the change, and make sure there's not any issue).

### LINKED ISSUE #43: Only fit `MinMaxScaler` on training data in LSTM examples
This comes from https://github.com/mlpack/models/pull/36#discussion_r371056033.

Right now, the LSTM examples use `MinMaxScaler` to scale the data.  But they use all of the data to fit the `MinMaxScaler`, which is actually a leakage of test data into the training process, and is not good data science practice. :)

So, it should be a simple change to ensure that `MinMaxScaler`'s `Fit()` method is only called with the training data.

kartikdutt18:
Hi everyone,
This closes #41, closes #42 and also closes #43.
I have used Ensmallen Callbacks instead of for loop to train LSTMs.
Another good idea that @shrit suggested is to use  mlpack's CLI functionality to improve access to models from command line.
This has been built and locally tested.
Thanks.

### DIFF
--- a/LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
+++ b/LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
@@ -203,36 +203,34 @@ int main()
   // we remove it also (first row in in arma::mat).
   dataset = dataset.submat(1, 1, dataset.n_rows - 1, dataset.n_cols - 1);
 
-  // Scale all data into the range (0, 1) for increased numerical stability.
-  data::MinMaxScaler scale;
-  scale.Fit(dataset);
-  scale.Transform(dataset, dataset);
-
   // We have 5 input data columns and 2 output columns (target).
   size_t inputSize = 5, outputSize = 2;
 
+  arma::mat trainData = dataset.submat(arma::span(),arma::span(0, (1 - RATIO) *
+      dataset.n_cols));
+  arma::mat testData = dataset.submat(arma::span(), arma::span((1 - RATIO) * dataset.n_cols,
+      dataset.n_cols - 1));
+
+  // Scale all data into the range (0, 1) for increased numerical stability.
+  data::MinMaxScaler scale;
+  // Fit scaler only on training data.
+  scale.Fit(trainData);
+  scale.Transform(trainData, trainData);
+  scale.Transform(testData, testData);
+
   // We need to represent the input data for RNN in an arma::cube (3D matrix).
-  // The 3rd dimension is rho, the number of past data records the RNN uses for.
+  // The 3rd dimension is the rho number of past data records the RNN uses for
   // learning.
-  arma::cube X, y;
-  X.set_size(inputSize, dataset.n_cols - rho + 1, rho);
-  y.set_size(outputSize, dataset.n_cols - rho + 1, rho);
-
-  // Create testing and training sets (read the notes above in the function
-  // definition).
-  CreateTimeSeriesData(dataset, X, y, rho);
-
-  // Split the data into training and testing sets.
   arma::cube trainX, trainY, testX, testY;
-  size_t trainingSize = (1 - RATIO) * X.n_cols;
-  trainX = X.subcube(arma::span(), arma::span(0, trainingSize - 1),
-      arma::span());
-  trainY = y.subcube(arma::span(), arma::span(0, trainingSize - 1),
-      arma::span());
-  testX = X.subcube(arma::span(), arma::span(trainingSize, X.n_cols - 1),
-      arma::span());
-  testY = y.subcube(arma::span(), arma::span(trainingSize, X.n_cols - 1),
-      arma::span());
+  trainX.set_size(inputSize, trainData.n_cols - rho + 1, rho);
+  trainY.set_size(outputSize, trainData.n_cols - rho + 1, rho);
+  testX.set_size(inputSize, testData.n_cols - rho + 1, rho);
+  testY.set_size(outputSize, testData.n_cols - rho + 1, rho);
+
+  // Create training sets for one-step-ahead regression.
+  CreateTimeSeriesData(trainData, trainX, trainY, rho);
+  // Create test sets for one-step-ahead regression.
+  CreateTimeSeriesData(testData, testX, testY, rho);
 
   // Only train the model if required.
   if (bTrain || bLoadAndTrain)
@@ -282,7 +280,7 @@ int main()
                 optimizer,
                 ens::PrintLoss(),
                 ens::ProgressBar(),
-                ens::EarlyStopAtMinLoss(),
+                ens::EarlyStopAtMinLoss(100),
                 ens::StoreBestCoordinates<arma::mat>());
 
     optimizer.ResetPolicy() = false;
--- a/LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
+++ b/LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
@@ -38,8 +38,8 @@ date  close  volume  open  high  low
 #include <mlpack/core/data/scaler_methods/min_max_scaler.hpp>
 #include <mlpack/methods/ann/init_rules/he_init.hpp>
 #include <mlpack/methods/ann/loss_functions/mean_squared_error.hpp>
+#include <mlpack/core/data/split_data.hpp>
 #include <ensmallen.hpp>
-#include <ensmallen_bits/callbacks/callbacks.hpp>
 
 using namespace std;
 using namespace mlpack;
@@ -49,16 +49,12 @@ using namespace ens;
 /**
  * Function to calcute MSE for arma::cube.
  */
-double MSE(arma::cube& pred, arma::cube& Y)
+/*
+ * Function to calcute MSE for arma::cube.
+ */
+double MSE(arma::cube &pred, arma::cube &Y)
 {
-  double err_sum = 0.0;
-  arma::cube diff = pred - Y;
-  for (size_t i = 0; i < diff.n_slices; i++)
-  {
-    arma::mat temp = diff.slice(i);
-    err_sum += accu(temp%temp);
-  }
-  return (err_sum) / (diff.n_elem + 1e-50);
+  return metric::SquaredEuclideanDistance::Evaluate(pred, Y) / (Y.n_elem);
 }
 
 /**
@@ -101,21 +97,13 @@ void SaveResults(const string filename,
                  const arma::cube& testX)
 {
   arma::mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
-  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
 
   // The prediction results are the (high, low) for the next day and come from
   // the last slice from the prediction.
-  arma::mat temp = predictions.slice(predictions.n_slices - 1);
+  flatDataAndPreds.rows(flatDataAndPreds.n_rows - 2,
+      flatDataAndPreds.n_rows - 1) = predictions.slice(predictions.n_slices - 1);
 
-  // We add 3 extra rows here in order to recreate the input data structure used
-  // to transform the data. This is needed in order to be able to use the right
-  // scaling parameters for the specific column stock high/low.
-  temp.insert_rows(0, 3, 0);
-  scale.InverseTransform(temp, temp);
-
-  // We add the prediction as the last two columns (stock high, low).
-  flatDataAndPreds.insert_rows(flatDataAndPreds.n_rows,
-      temp.rows(temp.n_rows - 2, temp.n_rows - 1));
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
 
   // We need to remove the last column because it was not used for training
   // (there is no next day to predict).
@@ -143,7 +131,7 @@ int main()
   // Change the names of these files as necessary. They should be correct
   // already, if your program's working directory contains the data and/or
   // model.
-  const string dataFile = "./../data/Google2016-2019.csv";
+  const string dataFile = "Google2016-2019.csv";
   // example: const string dataFile =
   //              "C:/mlpack-model-app/Google2016-2019.csv";
   // example: const string dataFile =
@@ -169,9 +157,6 @@ int main()
 
   // Testing data is taken from the dataset in this ratio.
   const double RATIO = 0.1;
-
-  // Number of iterations per cycle.
-  const int ITERATIONS_PER_EPOCH = 1000;
 
   // Step size of an optimizer.
   const double STEP_SIZE = 5e-5;
@@ -201,15 +186,18 @@ int main()
   // representation it is the first column.
   // The first column in the CSV is the date which is not required, therefore
   // we remove it also (first row in in arma::mat).
+
   dataset = dataset.submat(1, 1, dataset.n_rows - 1, dataset.n_cols - 1);
 
   // We have 5 input data columns and 2 output columns (target).
   size_t inputSize = 5, outputSize = 2;
 
-  arma::mat trainData = dataset.submat(arma::span(),arma::span(0, (1 - RATIO) *
-      dataset.n_cols));
-  arma::mat testData = dataset.submat(arma::span(), arma::span((1 - RATIO) * dataset.n_cols,
-      dataset.n_cols - 1));
+  // Split the dataset into training and validation sets.
+  arma::mat trainData, testData;
+  data::Split(dataset, trainData, testData, RATIO);
+
+  // Number of epochs for training.
+  const int EPOCHS = 150;
 
   // Scale all data into the range (0, 1) for increased numerical stability.
   data::MinMaxScaler scale;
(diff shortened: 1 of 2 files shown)


kartikdutt18:
LSTM Univariate Results:
```
Training : loss: 0.00605504
Finished training.
Saving Model
Model saved in lstm_univar.bin
Loading model ...
Mean Squared Error on Prediction data points: 0.00521374
The predicted energy consumption for the next hour is : 
 0.383799
```

kartikdutt18:
After sampling save results function, The new results are:
```
Mean Squared Error on Prediction data points: 0.00525148
The predicted energy consumption for the next hour is : 
 0.383751
```

and 

```
The predicted Google stock (high, low) for the last day is: 
  (956.784, 934.122)
```

These are similar to earlier values.

kartikdutt18:
Hi @zoq, Could you take a look at this. I think this is ready. I have repeated the tests twice to ensure that this works.

LSTM/TimeSeries-Univariate/src/LSTMTimeSeriesUnivariate.cpp
```diff
@@ -125,15 +122,15 @@ void SaveResults(const string& filename,
...
+      flatDataAndPreds.n_cols - 1) << endl;
 }
 
 int main()
 {
   // Change the names of these files as necessary. They should be correct
   // already, if your program's working directory contains the data and/or
   // model.
-  const string dataFile = "electricity-usage.csv";
+  const string dataFile = "./../data/electricity-usage.csv";
```
rcurtin: We should probably leave this as-is, to keep the example simple. :+1:

LSTM/TimeSeries-Univariate/src/LSTMTimeSeriesUnivariate.cpp
```diff
@@ -202,32 +193,34 @@ int main()
   // removing it also (first row in in arma::mat).
   dataset = dataset.submat(1, 1, 1, dataset.n_cols - 1);
 
+  arma::mat trainData = dataset.submat(arma::span(),arma::span(0, (1 - RATIO) *
+      dataset.n_cols));
+  arma::mat testData = dataset.submat(arma::span(), arma::span((1 - RATIO) * dataset.n_cols,
+      dataset.n_cols - 1));
+
+  // Number of iterations per cycle.
+  const int ITERATIONS_PER_EPOCH = trainData.n_cols * 300;
```
rcurtin: The `PER_EPOCH` in the name here is a bit confusing, because this now represents the total number of iterations.

Maybe a better idea might be to just define `EPOCHS` at the top, and then when constructing `optimizer`, just use `EPOCHS * trainData.n_cols`?

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -39,6 +39,7 @@ date  close  volume  open  high  low
 #include <mlpack/methods/ann/init_rules/he_init.hpp>
 #include <mlpack/methods/ann/loss_functions/mean_squared_error.hpp>
 #include <ensmallen.hpp>
+#include <ensmallen_bits/callbacks/callbacks.hpp>
```
rcurtin: I think this is included by default in `ensmallen.hpp`... correct me if I'm wrong. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -273,29 +262,22 @@ int main()
...
-      optimizer.ResetPolicy() = false;
+    cout << "Training ..." << endl;
 
-      arma::cube predOut;
-      // Getting predictions on test data points.
-      model.Predict(testX, predOut);
+    model.Train(trainX,
+                trainY,
+                optimizer,
+                ens::PrintLoss(),
```
rcurtin: It might be nice to add a comment above each callback indicating what it will do, just so that people coming through this code can understand why each one is used. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -273,29 +262,22 @@ int main()
...
+                ens::ProgressBar(),
+                ens::EarlyStopAtMinLoss(),
+                ens::StoreBestCoordinates<arma::mat>());
```
rcurtin: Does `StoreBestCoordinates` do anything here?  It looks like you are creating a temporary `StoreBestCoordinates()` object to pass to the optimizer, but then there is no way to get the result out since it is temporary.  Take a look at the ensmallen callback documentation; I think it has a nice example of using an instantiated `StoreBestCoordinates` callback.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -273,29 +262,22 @@ int main()
...
 
-      // Calculating mse on test data points.
-      double testMSE = MSE(predOut, testY);
-      cout << i + 1 << " - Mean Squared Error := " << testMSE << endl;
-    }
+    optimizer.ResetPolicy() = false;
```
rcurtin: Since we're only calling the optimizer once, this line is probably unnecessary. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
kartikdutt18: Agreed it is included. Sorry for my ignorance.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
kartikdutt18: Thanks for pointing it out. I got a chance to take a closer look at optimisers in ensmallen.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
kartikdutt18: Hmm, I removing it for now, I think the issue doesn't warrant `StoreBestCoordinates()` function however I think I need to use it locally to understand it a bit more clearly.

kartikdutt18:
Hi @rcurtin, Thanks for the review. I have made changes that you suggested.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: No worries, thanks for handling it. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -203,38 +189,40 @@ int main()
...
-  arma::cube X, y;
-  X.set_size(inputSize, dataset.n_cols - rho + 1, rho);
-  y.set_size(outputSize, dataset.n_cols - rho + 1, rho);
+  arma::mat trainData = dataset.submat(arma::span(),arma::span(0, (1 - RATIO) *
+      dataset.n_cols));
+  arma::mat testData = dataset.submat(arma::span(), arma::span((1 - RATIO) * dataset.n_cols,
+      dataset.n_cols - 1));
+
+  // Number of epochs for training.
+  const int EPOCHS = 500;
```
rcurtin: So, just a minor note, and whether or not you handle it is up to you.  Previously, we were doing 500 epochs, but they weren't really epochs because we were only looking at 1000 points in each "epoch".  Now we're using the whole dataset, and since one epoch now is equivalent to an entire pass (not just 1000 points), 500 epochs is a lot more training.  I wonder if maybe 100 is a better number here. :+1:

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: Sounds good, I think it's fine with or without `StoreBestCoordinates()`.

LSTM/TimeSeries-Univariate/src/LSTMTimeSeriesUnivariate.cpp
```diff
@@ -256,31 +249,26 @@ int main()
...
-    }
+    model.Train(trainX,
+                trainY,
+                optimizer,
+                // PrintLoss Callback prints loss for each epoch.
+                ens::PrintLoss(),
+                // Progressbar Callback prints progress bar for each epoch.
+                ens::ProgressBar(),
+                // Stops the optimization process if the loss stops decreasing
+                // or no improvement has been made. Useful in preventing overfitting.
```
rcurtin: This is a really pedantic comment but I think it's worth making---actually this is not useful for preventing overfitting in the way that people might expect.  Normally you might terminate the optimization when the loss on a *validation set* starts increasing (but the training set loss will keep going down).  However, `EarlyStopAtMinLoss()` only considers the training set.  So basically `EarlyStopAtMinLoss()` will terminate if the *training* error starts going up.  In my opinion the argument to say this is useful in preventing overfitting is a little weak, and I think a more accurate thing to say might be that this will terminate the optimization once we hit a minimum.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
kartikdutt18: I can make changes but I would have to test it first. Once I do I'll post a comment for the same. I hope that's okay.

LSTM/TimeSeries-Univariate/src/LSTMTimeSeriesUnivariate.cpp
```diff
@@ -256,31 +249,26 @@ int main()
...
-    // Run EPOCH number of cycles for optimizing the solution.
-    for (size_t i = 0; i < EPOCH; i++)
-    {
-      // Train neural network. If this is the first iteration, weights are
-      // random, using current values as starting point otherwise.
-      model.Train(trainX, trainY, optimizer);
-
-      // Don't reset optimizer's parameters between cycles.
-      optimizer.ResetPolicy() = false;
+    // Use Early Stopping criteria to stop training.
```
rcurtin: Another pedantic comment---this line itself doesn't specify that early stopping should be used, that's the callback.  I think a more effective comment might be:

```
// Instead of terminating based on the tolerance of the objective function, we'll depend on
// the maximum number of iterations, and terminate early using the EarlyStopAtMinLoss
// callback.
```

(I didn't check if those are 80 characters, so you might need to reflow it if you want to use that text directly.  Feel free to adapt it if you want to improve the wording or anything. :+1:)

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: Sure, no problem.  If 500 epochs are actually necessary we can go with that, I just figured it would generally terminate way before then.

LSTM/TimeSeries-Univariate/src/LSTMTimeSeriesUnivariate.cpp
kartikdutt18: Hmm, You are absolutely right, This will only get to a minima on training data. Nice I never looked at it that way. Thanks.

LSTM/TimeSeries-Univariate/src/LSTMTimeSeriesUnivariate.cpp
kartikdutt18: Great, I will make the changes for the same.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
kartikdutt18: On my laptop, it took 107 epochs to stop. So I have changed it to 150 (A nice round number). If needed I can reduce it further. Thanks for the suggestion.

kartikdutt18:
Hi @rcurtin, Since I made some changes that you suggested I tested it again and I got the following results:
```
Model saved in lstm_multi.bin
Loading model ...
Mean Squared Error on Prediction data points:= 0.116839
The predicted Google stock (high, low) for the last day is: 
  (1000.61, 978.512)
```
Similarly for Univariate results were:

```
Mean Squared Error on Prediction data points: 0.00492178
The predicted energy consumption for the next hour is : 
0.38489

```

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -57,7 +58,7 @@ double MSE(arma::cube& pred, arma::cube& Y)
     arma::mat temp = diff.slice(i);
     err_sum += accu(temp%temp);
   }
-  return (err_sum) / (diff.n_elem + 1e-50);
```
kartikdutt18: Hi @rcurtin, Is there a reason why we need to do this? 
I don't think diff.n_elem can ever be zero?

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
```diff
@@ -203,38 +190,39 @@ int main()
...
 
-  // We need to represent the input data for RNN in an arma::cube (3D matrix).
-  // The 3rd dimension is rho, the number of past data records the RNN uses for.
-  // learning.
-  arma::cube X, y;
-  X.set_size(inputSize, dataset.n_cols - rho + 1, rho);
-  y.set_size(outputSize, dataset.n_cols - rho + 1, rho);
+  // Split the dataset into training and validation sets.
+  arma::mat trainData, testData;
+  data::Split(dataset, trainData, testData, RATIO);
```
kartikdutt18: I have also switched to `Split` function from mlpack. I hope that's okay.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
kartikdutt18: `mlpack` has  `L2` distance in `core/metric`, would it be okay to switch to that. I think making our own MSE is a bit redundant. What do you think?

rcurtin: You're right, this is a lot cleaner.  Thanks! :+1:

rcurtin: And yeah, I agree, `diff.n_elem` should never be 0, so the `+ 1e-50` shouldn't ever be necessary.

LSTM/TimeSeries-Multivariate/src/LSTMTimeSeriesMultivariate.cpp
rcurtin: Hmm, actually I'm not sure on this one.  The `Split()` function shuffles the data during splitting, but for a time series problem like this it makes more sense to keep the first part of the data as the training set, and then the last part of the data as the test set.  Unfortunately I don't see any option to pass to `Split()` to avoid shuffling (maybe it would be good to add one?  I don't know how important it would be), so I think we should revert this bit, and then I can go ahead and merge it. :+1:

kartikdutt18: Yes you are right, I recently learned about this when I went through the code base for data for creating a DataLoader. This should definitely be reverted. Thanks for pointing it out.

kartikdutt18: I have made the changes for the same and tested it for epochs. Thanks for all the help.

rcurtin: Awesome, thanks.  I'll go ahead and merge it once it passes the tests. :+1:

kartikdutt18: Could also help me in answering this question, I was creating a DataLoader, So for time series analysis since input is transformed into output, does it make sense to fit only on training input and not all training data?
We would however transform both training input and output.

kartikdutt18: >Unfortunately I don't see any option to pass to `Split()` to avoid shuffling (maybe it would be good to add one? I don't know how important it would be)
I think optional shuffle will be a good idea. Should be simple to resolve this. Would it be okay for me to open a PR for this?

kartikdutt18:
Hmm, I think this is test failure is related to cmake changes added in mlpack and unrelated to models repo. I might be missing something though. I'll take a closer look just in case it is related to this PR.
Thanks.

rcurtin:
Yes, I think that due to mlpack#2247, the patch `mlpack-cmake.patch` is no longer necessary.  I *think* the build might succeed if you remove that file and remove the associated line from `.appveyor.yml`.  Do you want to try and see what happens? :)

rcurtin:
Oh, wait, sorry, I didn't see #58 which solves that issue.  Maybe we can just merge that, rebase this branch, and it should work. :)

kartikdutt18:
> Oh, wait, sorry, I didn't see #58 which solves that issue. Maybe we can just merge that, rebase this branch, and it should work. :)

Agreed, That would be great. Thanks.

kartikdutt18:
Hi, I have rebased this. I this should pass the tests now. Thanks.

birm:
I think this fits well with the other changes from models->examples, or at least doesn't conflict with them. If we're reworking LTSM, I think it would make more sense to do so after this is merged, rather than before.

kartikdutt18:
Agreed, That makes sense. This is ready from my side. Thanks a lot.

rcurtin:
Oops, I didn't realize that this wasn't merged before the repository split!  So I merged it here then cherry-picked the commits into the `models` repository.

kartikdutt18:
No worries, I think the models repo would be restructured with mlpack/models#3, so I'll try to take of anything that is needed there. Thanks a lot @rcurtin, @birm for the helpful reviews and comments.

PR #50, opened 2020-02-09
Adding AlexNet

kartikdutt18:
Hi everyone, I was working on adding alexnet to models however mlpack currently doesn't have adaptive pooling implemented. So I stopped here, the following things are left to be done:

- [x] Add AlexNet.
- [x] Write cmake to include alexnet as `#include<models/alexnet>`
- [ ] Addition of pre-trained weights.
- [ ] Changes in readme to show how models such as this one can be trained on their dataset.

**Currently Working on writing Cmake to resolve task 2.**

I hope this is ok.
Thanks.

PR #60, opened 2020-03-09
Restructuring code for models repo.

### LINKED ISSUE #40: Addition of MobileNet and architecture to load certain datasets.
Hi  I wanted to contribute to this organization by adding mobilenet (implementation similar to pytorch/tensorflow with attributes such as include top).
2. Secondly, I also wanted to add models folder where models can be stored as classes and a user can directly import them and train on their dataset.
3. A Samples folder where all models would have a test script where they can be trained directly.
4. A data-loader for standard datasets such as CIFAR 10, ImageNet (Currently only MNIST [From Kaggle] is available).
The implementation would result in model project to similar directory structure as other python ML/DL libraries where user could directly import a model, specify input size, number of classes, include top etc.) making using MLPack models very easy on any dataset.
To start off, I wanted to implement mobilenet, as it has a small architecture (less parameters) compared to other architectures such as ResNets and hence might be easier to train on CPU only.
Would it be okay if I pursue this issue?

### LINKED ISSUE #57: Implementation of Tests?
Hi everyone, This is in reference to PR #50, if models are added as libraries rather than cpp file so that they can be included in other cpp files, Rather than testing only on the system of the person who created the PR, does it makes sense to add a single pass on given dataset?
Another test would be loading of weights for the given model.
I am not sure if this needed but I think if changes are made in mlpack that might affect models here, a testing unit can help make the changes.
This will also prevent addition of models that have bugs. In the long run, this might also streamline addition of models.
I would love to hear your opinion.
Thanks.

kartikdutt18:
Hi everyone, Here is the list of current changes:

1. Implemented models (DigitRecognizer and LeNet as classes that can be included.)
2. Added DataLoader (Currently supports only mnist).
3. An examples folder to for mnist called mint_tutorial. 
4. Mnist dataset can be downloaded using `-DDOWNLOAD_MNIST` or `-DDOWNLOAD_DATASETS`
5. All examples use ensamllen.
6.  Unified Utils folder (Currently same as Kaggle_utils, Need cleaning up before I add LSTM.

Also closes #40, right now it partially handles #61, and aims to resolve #57.
I hope it would be okay to alexnet here as well.
I'll also add support for augmentation, Better utils, weights and test in the next 1 / 2 days. I would really appreciate your feedback.

Future Issues:
Added Augmentation.
Better DataLoader (support for selecting start and end columns)

Hi @zoq, @favre49, if you get a chance could please review this. I'll be making more changes to download weights and tests soon.


kartikdutt18:
I have also tested LeNet and simpleNN for 3 epochs each. They worked fine on machine generating 89% and 92% accuracy respectively on validation dataset.

prince776:
Hi @kartikdutt18 I think the implementation is really good. I checked the LeNet code and I have one suggestion, that is to add the option of `includeTop`. I'm unsure if it's needed in LeNet, but it's needed in models like VGG. So it will make models architecture consistent.
Let me know what you think about it.
I'm also looking forward to adding Inception to it.

kartikdutt18:
Hmm, I think the purpose of LeNet's paper was fully connected layer at the after Convolutions so if I remove it doesn't become a LeNet anymore however I don't mind adding that to be more consistent with the code base.

kartikdutt18:
Also @prince776, I probably skipped around some comments so if you notice something, could you review it and leave a comment saying that we should add this line here or something along similar lines. Thanks.

prince776:
> Also @prince776, I probably skipped around some comments so if you notice something, could you review it and leave a comment saying that we should add this line here or something along similar lines. Thanks.

Ok, I'll do that.

kartikdutt18:
Hi @prince776, Thanks for the review, I made constant changes to the code without changing comments so this is what we ended up with. Will fix them in the next commit. I hope you could take a look then too. Thanks.

ojhalakshya:
Hi @kartikdutt18 
Could you add more details about these classes.
Thanks.

models/simple_nn/simple_nn.hpp
```diff
@@ -0,0 +1,89 @@
...
+  //! Load weights into the model.
+  Sequential<>* LoadModel(const std::string &filePath);
+
+  //! Save weights for the model.
+  void SaveModel(const std::string& filePath);
+
+ private:
+  //! Locally stored SimpleNN Model.
+   Sequential<>* model;
+
```
ojhalakshya: Hi @kartikdutt18 
Can you look if there are 2 spaces required instead of 3?
Thanks.

prince776:
> Hi @prince776, Thanks for the review, I made constant changes to the code without changing comments so this is what we ended up with. Will fix them in the next commit. I hope you could take a look then too. Thanks.

Yes sure. I'm really excited to work on it anyways. I'll keep reviewing to make sure I'm up to date with the codebase.

models/simple_nn/simple_nn.hpp
kartikdutt18: Thanks for pointing it out. I kind of need many hands on deck with this. Could you also take a look at my future commits?
Thanks.

kartikdutt18:
Great, I don't think I'll be able to push a commit today. I am going with a complete redesign of DataLoader so I'll have to make changes for it accordingly. In the next commit I'll have restricted LSTMs, removed csv files completely from the repo and a better DataLoader.

kartikdutt18:
> Hi @kartikdutt18
> Could you add more details about these classes.
> Thanks.

Will do. Some of the portion of the code needs to heavily documented. Thanks.

kartikdutt18:
I haven't tested these changes, they probably require some debugging, but I spent yesterday making a DataLoader that ended up deleting so I made some changes to make progress. I'll post the result tomorrow after the fixes. Hopefully I'll also transfer VAE, and add tests tomorrow.

kartikdutt18:
Adding a todo list here to keep track of changes:

- [x] Models should be included as libraries / classes (sample can be seen in PR #50). So a user can call models/alexnet.hpp and use alexnet as base network for something else.

- [x] DataLoaders Folder (Will be pushing changes for it this week in PR#50). Here we should provide a wrapper around data::Load, scalers and data::Split. Especially useful for known datasets such as MNIST. This will also include support for resizing etc.
 
- [ ] A unified samples or as @favre49 suggested an examples folder (Refer #50) , where a call to each class should be explained. Eg. Object Detection Sample that has sample code for training and validation on a dataset. Similar file for object localization and so on.

- [ ] A testing folder (this will be for contributors) where they will add test to load weights train for 5 epochs and set a reasonable accuracy as baseline. This will ensure addition of valid models as well as streamline addition of models.

- [x] All data in one folder. I would however prefer if we made changes in cmake to download data rather than storing it in repo to make models repo lighter.

- [ ] Weights folder in models folder where weights of each models is stored.

kartikdutt18:
Hey @zoq, Could you please review the DataLoader, It's not the best representation right now, I wanted to know what would be the appropriate way of doing this such that more datasets can be added without hassle.

dataloaders/dataloader.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+  //! Locally stored Valid CSV.
+  arma::mat testCSVData;
+
+  // MNIST Dataset Dataloader.
+  void MNISTDataLoader();
+
+  // Google Stock Prices Dataloader.
+  void GoogleStockPricesDataloader();
+
```
jeffin143: I am not sure, why 3 different loaders, Can't we make one generic, Sorry I might have missed something

dataloaders/dataloader_impl.hpp
```diff
@@ -0,0 +1,140 @@
...
+    trainY = dataset.row(0) + 1;
+  }
+  else
+  {
+    data::Load("./../data/mnist_train.csv", dataset, true);
+    arma::mat train, valid;
+    data::Split(dataset, train, valid, ratio);
+
+    trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1) / 255.0;
+    validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1) / 255.0;
```
jeffin143: you can use mlpack internal scaler here, I guess

dataloaders/dataloader.hpp
```diff
@@ -0,0 +1,151 @@
...
+  //! Modify the Validation Dataset.
+  DataSetX &ValidX() { return validX; }
+
+  //! Get the Validation Dataset.
+  DataSetY ValidY() const { return validY; }
+  //! Modify the Validation Dataset.
+  DataSetY &ValidY() { return validY; }
+
+  //! Get the Training Dataset.
+  arma::mat TrainCSVData() const { return trainCSVData; }
```
jeffin143: How is it different from the follwing in line 62
```  
//! Modify the Training Dataset.
  DataSetX &TrainX() { return trainX; }```

Comment (kartikdutt18) [reply dataloaders/dataloader.hpp]: Ohk, so if user decides to store augmented data without splitting it into X and Y, he can access it using this. TrainX and Train Y are useful for classification only.

File: dataloaders/dataloader.hpp (hunk 2 above)
Comment (kartikdutt18) [reply dataloaders/dataloader.hpp]: We need at least two data loaders, one for Timeseries and other for datasets like CSV, I am working on implementing a better data loader as mentioned above. If you have any suggestions that would be great. Currently I'm thinking of implementing LoadTrainCSV and LoadTestCV which will be overloaded to support all CSV related datasets.

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (kartikdutt18) [reply dataloaders/dataloader_impl.hpp]: Hmm, then we would also need to store the scaler and then this would have to be done for all other datasets and that would just reduce flexibility. For Mnist we can use this. For other data loaders I have getters / setters to scale the data so user could call them to Fit and transform the data.

File: dataloaders/dataloader.hpp (hunk 4 above)
Comment (jeffin143) [reply dataloaders/dataloader.hpp]: Why can't we return them using a join_row(), joining TrainX and TrainY , why a special member ?

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (jeffin143) [reply dataloaders/dataloader_impl.hpp]: I am not sure if that would reduce flexibility, I guess that will allow user with more features since he can opt in for scaling.

File: dataloaders/dataloader.hpp (hunk 4 above)
Comment (kartikdutt18) [reply dataloaders/dataloader.hpp]: I am not sure it would work for cubes, I can try it though, but yes I think the dataloader needs to be improved a lot. This is currently in a working state only for now.

File: dataloaders/dataloader.hpp (hunk 2 above)
Comment (jeffin143) [reply dataloaders/dataloader.hpp]: Sorry, but can you briefly explain me why do we need at least two Data loaders ?

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (kartikdutt18) [reply dataloaders/dataloader_impl.hpp]: Hmm, I will keep that in mind while making a better data loader. The current data loader allowed me to run some tests since there a lot of changes. I'll push a commit with a better data loader and you could take a look at that.

File: dataloaders/dataloader.hpp (hunk 2 above)
Comment (kartikdutt18) [reply dataloaders/dataloader.hpp]: So time series dataset would need a parameter rho (look back) and modify data to hold past values, we would have to store data in cubes where as for other CSV datasets matrices are sufficient. So what I was planning to implement was we could have a LoadTrainCSV overloaded so that we can take rho if needed and store data in cubes. Here There would also be no shuffling of data. From this we can call the other LoadTrainCSV to perform tasks that would be similar. I might end up with 1 data loader also but I think I would know for sure only after I start making the changes.
I am sorry, I know this isn't really coherent but I think after I code it I can much better sense. I'll try to reduce redundant code as soon as possible.

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (jeffin143) [reply dataloaders/dataloader_impl.hpp]: If you don't mind can I work on your branch

File: dataloaders/dataloader.hpp (hunk 4 above)
Comment (jeffin143) [reply dataloaders/dataloader.hpp]: I will take a look

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (kartikdutt18) [reply dataloaders/dataloader_impl.hpp]: Sure, just one thing I need to mention after my last commit, I get the errors of either segfault or boost::failed assert. I would probably need to fix that first, then I could work on the data loader and I would love to get your help if you get a chance. Thanks.

File: dataloaders/dataloader.hpp (hunk 2 above)
Comment (jeffin143) [reply dataloaders/dataloader.hpp]: No issues, with it :). I am on it totally, I was just narrowing down the implementation.

We can call this v1, and I hope v2 and v3 will be much better 😉

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (kartikdutt18) [reply dataloaders/dataloader_impl.hpp]: I think for the plan you can refer to this [comment](https://github.com/mlpack/models/pull/60#issuecomment-598306692). I this is would be a very important PR for all future models that are added as we need to clearly lay down how a model, a dataset is to be added so it also ends up streamlining the flow. Much like mlpack, where we have a clear set of rules that we need to follow to add a feature hopefully through this PR we could achieve the same for this repo. Thanks.

Comment (jeffin143) [reply dataloaders/dataloader_impl.hpp]: > Sure, just one thing I need to mention after my last commit, I get the errors of either segfault or boost::failed assert. I would probably need to fix that first, then I could work on the data loader and I would love to get your help if you get a chance. Thanks.

I really liked the idea, It was on my todo list, Something of transfer learning and the whole idea. I will take a look over this coming weekend

File: dataloaders/dataloader.hpp (hunk 2 above)
Comment (kartikdutt18) [reply dataloaders/dataloader.hpp]: I will try my best and hopefully v2 will be much better than what it currently is. 👍

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (kartikdutt18) [reply dataloaders/dataloader_impl.hpp]: I agree, I think this would be very useful especially for training large models. We could train without top (Add it outside model class) and store the weights. Then we could use these weights in a larger model like YOLO. This should decrease training epochs required hence reduce training time.

Comment (kartikdutt18) [reply dataloaders/dataloader_impl.hpp]: Something similar to ladder training.

**Comment by kartikdutt18** [comment]
Hey @jeffin143, any suggestions for CLI?

File: dataloaders/dataloader_impl.hpp (hunk 3 above)
Comment (kartikdutt18) [reply dataloaders/dataloader_impl.hpp]: Yayy, So I fixed the build, testing on different OS now.

File: dataloaders/dataloader.hpp [hunk 5]
```diff
@@ -0,0 +1,151 @@
...
+
+  //! Locally stored ratio for train-test split.
+  double ratio;
+
+  //! Locally stored augmentation.
+  std::vector<std::string> augmentation;
+
+  //! Locally stored augmented probability.
+  double augmentationProbability;
+};
```
prince776: a comment like  // class DataLoader should be added here after  `};`.

dataloaders/dataloader.hpp
```diff
@@ -0,0 +1,151 @@
...
+#include <mlpack/prereqs.hpp>
+
+#ifndef MODELS_DATALOADER_HPP
+#define MODELS_DATALOADER_HPP
+
+template<
+  typename DataSetX = arma::mat,
+  typename DataSetY = arma::mat
+>
+class DataLoader
```
prince776: I think we should add all these classes within a namespace. like `mlpack::dataloaders`.

examples/time_series_tutorial/electricity_consumption_tutorial.cpp
```diff
@@ -0,0 +1,75 @@
...
+              optimizer,
+              ens::PrintLoss(),   // PrintLoss Callback prints loss for each epoch.
+              ens::ProgressBar(), // Progressbar Callback prints progress bar for each epoch.
+              ens::EarlyStopAtMinLoss());
+
+  cout << "Validation Results: \n MSE:" << std::endl;
+  arma::cube preds;
+  model.Predict(dataloader.TestX(), preds);
+  cout << MSE(preds, dataloader.TestY()) << endl;
+}
```
prince776: extra line at eof needed.

kartikdutt18:
Thanks @prince776, I will make the changes. Also there is a typo in the data loader that I will fix in the next commit, I spent hours using valgrind to see why it segfaults.

dataloaders/dataloader_impl.hpp
```diff
@@ -0,0 +1,139 @@
...
+>::GoogleStockPricesDataloader()
+{
+  arma::mat dataset;
+  data::Load("./../data/Google2016-2019.csv", dataset, true);
+  dataset = dataset.submat(1, 1, dataset.n_rows - 1, dataset.n_cols - 1);
+
+  trainCSVData = dataset.submat(arma::span(),arma::span(0, (1 - ratio) *
+       dataset.n_cols));
+  testCSVData = dataset.submat(arma::span(), arma::span((1 - ratio) * dataset.n_cols,
+       dataset.n_cols - 1));
```
jeffin143: data::split() can be used here, without shuffling

kartikdutt18: Yay!! Thanks for the PR. I will hopefully push a new data loader by tonight.

jeffin143: Raised a [#2293](https://github.com/mlpack/mlpack/pull/2293)

kartikdutt18:
Hey @jeffin143, sorry to bother you again but do you think the new data loader is better that what it initially was?
I haven't added comments yet, I also have to add support for google-stock-prices dataset but in general does seem like a better idea?

dataloaders/dataloader.hpp
```diff
@@ -0,0 +1,195 @@
...
+                   const int endInputFeatures = -1);
+
+  void LoadTestCSV(const std::string &datasetPath,
+                   const int rho = 10,
+                   const bool useScaler = false,
+                   const bool dropHeader = false,
+                   const int startInputFeatures = -1,
+                   const int endInputFeatures = -1,
+                   const size_t inputSize = 1,
+                   const size_t outputSize = 1);
```
jeffin143: @kartikdutt18 , I just took a small glance, I guess we don't need all these  mutiple functions

May be just use `LoadCsv()` function 

And the if user wants he can call this function two times 
`LoadCSv(input, "mytest")` -> For testCSV
`LoadCsv(input_train, "Mytrain")` -> for trainCSV

or if he isn't having the data in splited format ,then he can call the function to load data and split it too 

`LoadCsv ( input, "data", split = True )` ForTestTrainCSV 

Not sure, This is a vague idea, I  am thinking aloud, And need probably to write down , I was currently working on something and hence did dwell much, Will take a look tonight

kartikdutt18: I can do that but I would end up adding if conditions which aren't clean but I do get your point and I don't mind doing that. I think I can add another option `time_series = true / false` To remove remaining overloaded functions.

models/lenet/lenet.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * @param s The stride size (x or y direction).
+   * @param padding The size of the padding (width or height) on one side.
+   * @return The convolution output size.
+   */
+  size_t ConvOutSize(const size_t size,
+                     const size_t k,
+                     const size_t s,
+                     const size_t padding)
+  {
+    return std::floor(size + 2 * padding - k) / s + 1;
```
prince776: Shouldn't this be `return std::floor((size + 2 * padding - k) / s )+ 1;`

models/lenet/lenet.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * @param size The size of the input (row or column).
+   * @param k The size of the filter (width or height).
+   * @param s The stride size (x or y direction).
+   * @return The convolution output size.
+   */
+  size_t PoolOutSize(const size_t size,
+                     const size_t k,
+                     const size_t s)
+  {
+    return std::floor(size - 1) / s + 1;
```
prince776: same here ` return std::floor((size - 1) / s) + 1;

models/lenet/lenet.hpp
kartikdutt18: The models works for me, So I'll change them after testing. Resolving this for now.

kartikdutt18: So I looked into it just now, it doesn't matter. C++ always floors division anyway so we remove it completely if we want.

kartikdutt18:
Closing this, It will be shifted to the new repo.

PR #67, opened 2020-03-21
Update readme.

rcurtin:
This rewrites the readme for this repository to focus on the fact that it's for simple examples.

I've removed all listing of models, since hopefully that should be pretty clear from the directory tree (if we name the examples accordingly).

Let me know what you think---happy to change some of the wording, etc. :+1:


rcurtin:
@birm good point.  I added a comment that not everything is done yet, and I'll go ahead and merge now so that we can keep things moving. :)

PR #69, opened 2020-03-26
Strip CMake from all examples and restructure each example into its own directory

rcurtin:
This is a pretty simple first pass at restructuring, and isn't meant to be final---my goal is just to get us to a place where it's easy to contribute to the restructuring effort further.  I've done the following:

 - Anything referencing CMake is now gone (in accordance with the goals of the repository).
 - Each example is in its own top-level directory.  This means that all utility code is also in the same directory.
 - I updated the README to include a bit about each example.

There are still a bunch of things to be done, which I'll open as issues and can be handled in the future:

 - Makefiles should be created for each example
 - The travis and AppVeyor configurations should be updated to build each example in the Makefile. (And make sure they run with no input?  Maybe that's worth further discussion)
 - All utility files should be removed; these examples should be as simple as possible.
 - Something should be done with the `vae_utils/` directory, but I'm not actually quite sure what.  We can figure it out later.

Let me know what you think. :+1:


mnist_cnn/kaggle_utils.hpp
```diff
@@ -0,0 +1,90 @@
+/**
+ * Utitlity functions that is useful for solving Kaggle problems
```
kartikdutt18: I think after #55, We would be able to remove this file as well.

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -46,9 +46,6 @@ using namespace mlpack;
 using namespace mlpack::ann;
 using namespace ens;
 
-/**
- * Function to calcute MSE for arma::cube.
- */
 /*
  * Function to calcute MSE for arma::cube.
```
favre49:
```suggestion
 * Function to calculate MSE for arma::cube.
```

mnist_cnn/kaggle_utils.hpp
shrit: Yes, I agree, maybe I am going to remove it from my pull request.

shrit:
@kartikdutt18 Considering the `vae` I think we can keep only the example here, all declaration, code related to the `vae` can be removed from here.

mnist_cnn/kaggle_utils.hpp
rcurtin: Ah, nice, if you want to put that in #55 @shrit that would be great, otherwise we can handle it after #55 is merged. :+1:

lstm_stock_prediction/lstm_stock_prediction.cpp
rcurtin: Thanks! :+1:

rcurtin:
Sure, VAE is a little complex for an example perhaps.  I've removed it in acdf50f. :+1:

mnist_cnn/kaggle_utils.hpp
kartikdutt18: > Yes, I agree, maybe I am going to remove it from my pull request.

That would be great. Thanks a lot.

kartikdutt18:
> @kartikdutt18 Considering the vae I think we can keep only the example here, all declaration, code related to the vae can be removed from here.

I might have understood, but I think you are suggesting that we keep something like `van_cnn.cpp` and remove rest of the files. I think this should be very simple, a single vae_cnn file and single vae file (here folder). I will try to see if that makes sense.

rcurtin:
@kartikdutt18 sounds good, I re-added the MNIST VAE CNN example.  If you think this is good to go, I'm happy to go ahead and merge it. :+1: (It's not meant to be perfect, just a start to organizing things.)

kartikdutt18:
Hey @rcurtin, I think this is only to set the general structure right?
I think after this I will try to simplify VAE as well to remove vae utils. I hope that's okay.

rcurtin:
@kartikdutt18 yep, exactly.  That sounds good to me.  The Travis build is "passing" now, so from my side I'm ready to merge.  Any other comments/thoughts or should I hit the button? :)

kartikdutt18:
I think we can merge this then. Thanks a lot for this. Do you mind taking a look at restructuring PR at models repo too (about the general structure and the code in general).

rcurtin:
I'll take a look when I have a chance---any reviews I do today I want to focus first on those that are related to the 3.3.0 release, but I'll see if I can get to that next (I consider the repo restructuring here high priority, at least until it's in a fairly stable state).

kartikdutt18:
I completely understand. Thanks a lot.

PR #74, opened 2020-04-04
Example notebooks

zoq:
Example notebooks, haven't used much of the Jupyter notebook functions for comments, so this could be directly copied into C++ or Python to run the code outside of a notebook.

Also, the badge at the top, will open a notebook instance to run the examples, but right now it will just open an empty instance, once we can agree this is a good idea I can automatically clone the example repository.


zoq:
Right now it runs on a machine that is basically unused (no load or traffic restriction). That said, I think at some point we can switch to https://mybinder.org/, once xeus-cling is updated. Also, the setup runs on docker swarm, so we can add more machines if we wanted.

mnist-cnn-cpp.ipynb
```diff
@@ -0,0 +1,569 @@
...
+   "cell_type": "code",
+   "execution_count": 26,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "/**\n",
+    " * Returns the accuracy (percentage of correct answers).\n",
+    " *\n",
+    " * @param predLabels predicted labels of data points.\n",
+    " * @param realY real labels (they are double because we usually read them from CSV file that contain many other double values).\n",
```
favre49, line 116: I think this line is too long? It doesn't display neatly on the rich diff.

In general, do we stick to the mlpack style guidelines here as well?

mnist-cnn-cpp.ipynb
```diff
@@ -0,0 +1,569 @@
...
+    "constexpr int BATCH_SIZE = 50;"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Labeled dataset that contains data for training is loaded from CSV file. Rows represent features, columns represent data points.\n",
```
favre49, line 198: This line as well

mnist-cnn-cpp.ipynb
```diff
@@ -0,0 +1,569 @@
...
+    "data::Load(\"mnist-train.csv\", tempDataset, true);"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// The original Kaggle dataset CSV file has headings for each column, so it's necessary to get rid of the first row. In Armadillo representation, this corresponds to the first column of our data matrix.\n",
```
favre49, line 211: This line as well

mnist-cnn-cpp.ipynb
zoq: From my side that was intentional, because jupyter and the rendered github output will break the line, so to save some space in the notebook itself I put everything on a single line. But I can revert to the mlpack style.

favre49: Oh alright, if it breaks the line properly I don't mind either way, I'll leave it up to you. In the rich diff it broke the ending word as well, which is why I made the comment.

covertype-rf-cpp.ipynb
```diff
@@ -0,0 +1,151 @@
...
+    "using namespace mlpack::tree;"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Load the dataset from an online URL.\n",
```
rcurtin, line 67: I guess technically it's not an online URL, since we downloaded with `wget` earlier.  (I know I'm being pedantic about that, sorry.)  Maybe `Load the dataset that we downloaded.`?

covertype-rf-cpp.ipynb
```diff
@@ -0,0 +1,151 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "arma::mat trainSet, testSet;\n",
+    "arma::Row<size_t> trainLabels, testLabels;\n",
+    "\n",
+    "data::Split(dataset, labels, trainSet, testSet, trainLabels, testLabels, 0.3 /* Percentage of dataset to use for test set. */);"
```
rcurtin, line 92: Given that this is a super simple example, it could be useful to add a comment just above this---`split dataset randomly into training set and test set`, or something.

covertype-rf-py.ipynb
```diff
@@ -0,0 +1,145 @@
...
+    "[![mlpack-lab Image](https://img.shields.io/endpoint?url=https%3A%2F%2Flab.kurg.org%2Fstatus%2Fstatus.json)](https://lab.mlpack.org)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# @file iris-pca-py.ipynb\n",
```
rcurtin, line 16:
```suggestion
    "# @file covertype-rf-py.ipynb\n",
```

iris-pca-cpp.ipynb
```diff
@@ -0,0 +1,164 @@
...
+      "   5.0000   3.4000   1.5000\n",
+      "   4.4000   2.9000   1.4000\n",
+      "   4.9000   3.1000   1.5000\n",
+      "   5.4000   3.7000   1.5000\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "// Print the first 10 rows of the input data.\n",
```
rcurtin, line 104: I think it would be useful to point out why we have to transpose (because Armadillo stores data in column-major format, so if we want to show a point as each row, we have to transpose before we print).  I imagine that for many people looking at these examples, they will have never seen mlpack and never looked at other documentation, so they may have no idea that mlpack is column-major internally via Armadillo.

iris-pca-py.ipynb
```diff
@@ -0,0 +1,137 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 19,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load the dataset from an online URL.\n",
+    "data = pd.read_csv('https://lab.mlpack.org/data/iris.csv')"
```
rcurtin, line 39: Maybe this needs `header=None`?  It looks like in the next cell that the first row is getting used as headers.  (I think `header=None` is the right option.)

iris-pca-py.ipynb
```diff
@@ -0,0 +1,137 @@
...
+    "# Print the first ten columns of the transformed input.\n",
+    "data['output'][1:10]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
```
rcurtin, line 113:
```suggestion
```
We can probably remove the empty cell.  I *think* this suggestion will do that correctly, but I haven't tested it. :)

mnist-cnn-cpp.ipynb
```diff
@@ -0,0 +1,569 @@
...
+   "cell_type": "code",
+   "execution_count": 41,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "tar: mnist-train.csv: Cannot utime: Operation not permitted\n",
+      "tar: Exiting with failure status due to previous errors\n"
```
rcurtin, line 37: Maybe we should rerun this to get correct output?

mnist-cnn-cpp.ipynb
rcurtin: The lines seem too long for my display:

notebook_too_wide

Maybe my browser is too wide?  I'm not sure the screen resolution you were testing with.

mnist-cnn-cpp.ipynb
```diff
@@ -0,0 +1,569 @@
...
+   "source": [
+    "/**\n",
+    " * @file mnist-cnn-cpp.ipynb\n",
+    " *\n",
+    " * An example of using Convolutional Neural Network (CNN) for\n",
+    " * solving Digit Recognizer problem from Kaggle website.\n",
+    " *\n",
+    " * The full description of a problem as well as datasets for training\n",
+    " * and testing are available here https://www.kaggle.com/c/digit-recognizer.\n",
+    " */"
```
rcurtin, line 24: This is a really awesome example!  I think the changes @shrit has made to this example in #55 could simplify the notebook, but we can make those changes to this notebook later.

mnist-cnn-cpp.ipynb
```diff
@@ -0,0 +1,569 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\u001b[0;33m[WARN ] \u001b[0mThe optimizer's maximum number of iterations is less than the size of the dataset; the optimizer will not pass over the entire dataset. To fix this, modify the maximum number of iterations to be at least equal to the number of points of your dataset (37800).\n",
```
rcurtin, line 374: After the changes from #55 the output here will get a lot nicer too, since it's correctly using callbacks. :)

mnist-cnn-cpp.ipynb
zoq: Maybe, if that looks strange on any of your systems, let's revert to the original style.

zoq:
> Just curious, why does it need access to the Github account? Is that a Jupyterlab restriction or something?

There are all sorts of authenticators: https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html, I wanted to have some sort of authentication, because that allows you to share/save notebooks as well and it is also somewhat an option to hinder someone to open hundreds and hundreds of instances.

zoq:
> I know that notebooks will let you save directly as a .cpp file---do you think we want to add those files to the repository too, in case someone doesn't want to use the notebook?

Sure, do you think it still makes sense to have everything in a single folder? In this PR I put everything in root, since there is no ipynb extension that reflects the kernel I used `-py` and `cpp` to distinguish between them.

mnist-cnn-cpp.ipynb
zoq: Yes, good point.

rcurtin:
> > I know that notebooks will let you save directly as a .cpp file---do you think we want to add those files to the repository too, in case someone doesn't want to use the notebook?
> 
> Sure, do you think it still makes sense to have everything in a single folder? In this PR I put everything in root, since there is no ipynb extension that reflects the kernel I used `-py` and `cpp` to distinguish between them.

Ah, personally I think "one example per directory" is probably the best way to go.  So, we could make a directory `forest_covertype_prediction_with_random_forests/` (I know it's a long name but it's descriptive; maybe you can think of something shorter but equally descriptive?), and then put `covertype-rf-cpp.ipynb`, `covertype-rf-py.ipynb`, and I guess `covertype-rf.cpp` and `covertype-rf.py` in the directory.  What do you think; does that sound reasonable?

zoq:
Yes, sounds like a good approach to me.

birm:
Is this ready, or do you still want to move the examples around?

zoq:
Yes, have to move the files, will see if I can get it done in the next days.

favre49:
Just one thing before we merge. In [this](https://github.com/mlpack/examples/pull/74#issuecomment-615995928) comment, @rcurtin said that each notebook should be in a single folder. Personally i find it intuitive to have every q-learning notebook in a single folder, but that's just me and we should follow one standard throughout the repository. Perhaps we should rename this folder?

rcurtin:
Let me know if there's anything I can do to help with this one---it's quite exciting and I would love to see it merged soon. :+1:

zoq:
Thanks, I'm switching to binder to make this simpler to maintain, once that is done I will update the PR.

zoq:
The last commits include multiple changes:

- Switch from a default Jupyterhub to a Binderhub installation, which makes it a lot easier to update the environment, because the config is part of the repository. As a first step Binderhub builds a custom docker image that includes all the necessary dependencies; including some mlpack specific modifications to reduce the burden to include Jupyter-kernel specific settings.
All that means we can use https://mybinder.org/ to run the notebooks. Since binder is free for everyone, the hardware specs are somewhat limited (2GB memory, limited CPU), so running the examples can take some time. Because of that, I went ahead and set up our own Binderhub installation https://lab.mlpack.org. Each link at the top of the notebook should use the lab.mlpack.org instance (once the code is merged).

- New kernels, the language bindings are growing fast, mlpack can now be used in Julia and Go, so I went ahead configured the kernels and added an example for each, I’m super impressed with the performance of the Go kernel, to me it feels a lot faster as the C++ kernel.

- I removed the other examples from the PR for now, I will open a new PR to reinclude them, but just to see if everything works as planned it’s probably easier to merge one example after another.

You can already play with the code here:


**lab.mlpack.org** -> https://lab.mlpack.org/v2/gh/zoq/examples/example-notebooks
**mybidner.org** -> https://mybinder.org/v2/gh/zoq/examples/example-notebooks

forest_covertype_prediction_with_random_forests/covertype-rf-py.ipynb
```diff
@@ -0,0 +1 @@
...test_labels = output['test_labels']","metadata":{"trusted":true},"execution_count":16,"outputs":[]},{"cell_type":"code","source":"# Train a random forest.\noutput = mlpack.random_forest(training=training_set, labels=training_labels,\n    print_training_accuracy=True, num_trees=10, minimum_leaf_size=3)","metadata":{"trusted":true},"execution_count":17,"outputs":[]},{"cell_type":"code","source":"random_forest = output['output_model']","metadata":{"trusted":true},"execution_count":18,"outputs":[]},{"cell_type":"code","source":"# Predict the labels of the test points.\noutput = mlpack.random_forest(input_model=random_forest, test=test_set)","metadata":{"trusted":true},"execution_count":19,"outputs":[]},{"cell_type":"code","source":"# Now print the accuracy. The 'probabilities' output could also be used to\n# generate an ROC curve.\ncorrect = np.sum(output['predictions'] == test_labels.flatten())\nprint(str(correct) + ' correct out of ' + str(len(test_labels)) +\n      ' (' + str(100 * float(correct) / float(len(test_labels))) + '%).')","metadata":{"trusted":true},"execution_count":20,"outputs":[{"name":"stdout","text":"24513 correct out of 30000 (81.71%).\n","output_type":"stream"}]}]}
```
birm, line 1: Do these need to be single line files?

zoq: Hm, downloaded straight from Jupyter-lab, but let me prettify the output.

binder/Project.toml
```diff
@@ -0,0 +1,12 @@
...
+author = ["mlpack developers <mlpack@lists.mlpack.org>"]
+version = "3.3.2"
+
+[deps]
+Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+mlpack_jll = "156a7db3-27ba-586e-a86b-f061da4f95ea"
+
+[compat]
+julia = "1.3"
+mlpack_jll = "3.3.2"
```
rcurtin, line 12: Are you sure this file is needed?  It looks like the `Pkg.add("mlpack")` in the Julia notebook should be sufficient.

forest_covertype_prediction_with_random_forests/covertype-rf-cpp.ipynb
```diff
@@ -0,0 +1,152 @@
...
+      "cell_type":"code",
+      "source":"// Now print the accuracy. The 'probabilities' output could also be used to\n// generate an ROC curve.\nconst size_t correct = arma::accu(output == testLabels);\nstd::cout << correct\n          << \" correct out of \"\n          << testLabels.n_elem << \" (\"\n          << 100.0 * correct / testLabels.n_elem\n          << \").\" << std::endl;",
+      "metadata":{
+        "trusted":true
+      },
+      "execution_count":13,
+      "outputs":[
+        {
+          "name":"stdout",
+          "text":"29798 correct out of 30000 (99.3267).\n",
```
rcurtin, line 146: When I ran this I only got something like 82% accuracy; maybe you ran it with different configuration here or something?

zoq: I used the same code/configuration, I guess I got a 'better' train/test split. I can rerun the code to get another output?

rcurtin: Wow, that's quite surprising that the difference is 17%... maybe it's worth trying again and seeing what you get?

binder/Project.toml
rcurtin: Ah, ok.  You might consider making an empty `Project.toml` file in a new directory with

```
julia> import Pkg
julia> Pkg.activate("mlpack-notebooks")
```

which (hopefully) should make a simple `Project.toml` for a project called `mlpack-notebooks`.  I suppose, it may also be possible that you could add `mlpack` to the list of package dependencies and maybe it will make the `Pkg.add()` lines in the notebook unnecessary.  I'm not 100% sure it will work, but maybe worth trying? :)

forest_covertype_prediction_with_random_forests/covertype-rf-cpp.ipynb
zoq: Agreed, updated the notebook output.

binder/Project.toml
zoq: Looks like if you provide an initial `Project.toml` in the root folder it will install the deps, so I removed `Pkg.add` from the notebook.

Project.toml
```diff
@@ -0,0 +1,8 @@
+[deps]
+CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+Libz = "2ec943e9-cfe8-584d-b93d-64dcb6d567b7"
+mlpack = "cf18a64e-e90e-11e8-37b7-fb5df6478bc0"
+
+[compat]
+julia = "1.3"
```
rcurtin, line 8: Is this needed in the root directory too?  (I'm not sure if it's extra.)

zoq: Yes, that's an issue with repo2docker, the `Project.toml` in the binder folder tells repo2docker to install the Julia kernel, but doesn't install the deps, the `Project.toml` in the root folder only installs the deps but not the kernel. Pretty sure this will get solved at some point, in which case we can one of the files.

rcurtin: Ahhh, got it.  Thanks for the clarification. :+1:

PR #75, opened 2020-04-07
Example of possible makefile for mlpack examples

shrit:

### DIFF
--- a/lstm_electricity_consumption/Makefile
+++ b/lstm_electricity_consumption/Makefile
@@ -13,7 +13,7 @@ CXXFLAGS += -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp
 # Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
 # CXXFLAGS += -std=c++11 -Wall -Wextra -g -O0
 LDFLAGS  += -fopenmp
-# LDFLAGS += -L/path/to/mlpack/library/    # if installed locally
+# LDFLAGS += -L/path/to/mlpack/library/    # If installed locally.
 INCLFLAGS := -I . #Add header directories for any includes that aren't on the default compiler search path.
 CXXFLAGS += $(INCLFLAGS)
 
--- a/lstm_electricity_consumption/Makefile
+++ b/lstm_electricity_consumption/Makefile
@@ -9,12 +9,13 @@ SRC := lstm_electricity_consumption.cpp
 LIBS_NAME := armadillo mlpack boost_serialization
 
 CXX := g++
-CXXFLAGS := -std=c++11 -Wall -Wextra -O3 -DNDEBUG
+CXXFLAGS += -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp
 # Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
-# CXXFLAGS := -std=c++11 -Wall -Wextra -g -O0
+# CXXFLAGS += -std=c++11 -Wall -Wextra -g -O0
 LDFLAGS  := -fopenmp
 # LDFLAGS := -L/path/to/mlpack/library/    # if installed locally
 INCLFLAGS := -I . #Add header directories for any includes that aren't on the default compiler search path.
+CXXFLAGS += $(INCLFLAGS)
 
 OBJS := $(SRC:.cpp=.o)
 LIBS := $(addprefix -l,$(LIBS_NAME))
@@ -24,7 +25,7 @@ CLEAN_LIST := $(TARGET) $(OBJS)
 default: all
 
 $(TARGET): $(OBJS)
-	$(CXX) $(CXXFLAGS) $(OBJS) -o $(TARGET) $(INCLFLAGS) $(LIBS) $(LDFLAGS)
+	$(CXX) $(CXXFLAGS) $(OBJS) -o $(TARGET) $(LDFLAGS) $(LIBS)
 
 .PHONY: all
 all: $(TARGET)
--- a/lstm_electricity_consumption/Makefile
+++ b/lstm_electricity_consumption/Makefile
@@ -10,10 +10,11 @@ LIBS_NAME := armadillo mlpack boost_serialization
 
 CXX := g++
 CXXFLAGS += -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp
-# Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
+# Use these CXXFLAGS instead if you want to compile with debugging symbols
+# and without optimizations.
 # CXXFLAGS += -std=c++11 -Wall -Wextra -g -O0
 LDFLAGS  += -fopenmp
-# LDFLAGS += -L/path/to/mlpack/library/    # If installed locally.
+LDFLAGS += -L . # /path/to/mlpack/library/  # If installed locally.
 # Add header directories for any includes that aren't on the default compiler
 # search path.
 INCLFLAGS := -I .
--- a/lstm_electricity_consumption/Makefile
+++ b/lstm_electricity_consumption/Makefile
@@ -14,7 +14,9 @@ CXXFLAGS += -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp
 # CXXFLAGS += -std=c++11 -Wall -Wextra -g -O0
 LDFLAGS  += -fopenmp
 # LDFLAGS += -L/path/to/mlpack/library/    # If installed locally.
-INCLFLAGS := -I . #Add header directories for any includes that aren't on the default compiler search path.
+# Add header directories for any includes that aren't on the default compiler
+# search path.
+INCLFLAGS := -I .
 CXXFLAGS += $(INCLFLAGS)
 
 OBJS := $(SRC:.cpp=.o)
--- a/lstm_electricity_consumption/Makefile
+++ b/lstm_electricity_consumption/Makefile
@@ -12,8 +12,8 @@ CXX := g++
 CXXFLAGS += -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp
 # Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
 # CXXFLAGS += -std=c++11 -Wall -Wextra -g -O0
-LDFLAGS  := -fopenmp
-# LDFLAGS := -L/path/to/mlpack/library/    # if installed locally
+LDFLAGS  += -fopenmp
+# LDFLAGS += -L/path/to/mlpack/library/    # if installed locally
 INCLFLAGS := -I . #Add header directories for any includes that aren't on the default compiler search path.
 CXXFLAGS += $(INCLFLAGS)
 
--- a/lstm_electricity_consumption/Makefile
+++ b/lstm_electricity_consumption/Makefile
@@ -1,0 +1,35 @@
+# This is a simple Makefile used to build the example source code.
+# This example might requires some modifications in order to work correctly on
+# your system.
+# If you're not using the Armadillo wrapper, replace `armadillo` with linker commands
+# for the BLAS and LAPACK libraries that you are using.
+
+TARGET := lstm_electricity_consumption
+SRC := lstm_electricity_consumption.cpp
+LIBS_NAME := armadillo mlpack boost_serialization
+
+CXX := g++
+CXXFLAGS := -std=c++11 -Wall -Wextra -O3 -DNDEBUG
+# Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
+# CXXFLAGS := -std=c++11 -Wall -Wextra -g -O0
+LDFLAGS  := -fopenmp
+# LDFLAGS := -L/path/to/mlpack/library/    # if installed locally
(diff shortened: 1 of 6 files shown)


shrit:
@rcurtin  I think this is a possible example for a Makefile. I only made it for one example, if it is good, I will copy it everywhere with minor modifications.

birm:
Maybe we could add this makefile to our CI tests.

zoq:
Do we want to keep the current folder structure or do we want to put everything in the root directory? Depending on the decision, I think a single Makefile as all we need?

shrit:
@zoq I would prefer to have separate `Makefile` by example. This will reduce the compilation time, and will be much easier to manage and debug. This is only a prototype, if you find it well I will only copy it for all examples.

zoq:
Maybe we can make it work for other examples? Like point users to this makefile and we make sure it works across all examples, and a user could call `make example-name` and that's it?

shrit:
@zoq I agree, that is a good point, this will reduce non-useful duplication of Makefiles. Let us have @rcurtin opinion on this?

rcurtin:
@shrit thanks for writing this up!  I think I would agree that we should have one Makefile in each directory.  My impression is that a user will come along and look at an individual directory, not at the whole project.  But it seems like it would be easy to have the best of both worlds... just have a top-level Makefile call out recursively to all the subdirectories?

This is a really generic Makefile, which is nice because we could just drop it in other directories, but it might be a little hard to understand for people not too familiar with the (really weird!) Makefile language.  In some sense even the Makefile itself is an example, so I'd hope to keep it as simple as we can.  But I'm not sure exactly where the right level is.  The configurable parameters at the top are really nice, but maybe we can hardcode the sources and compilation commands, so we get something like this?

```
all: mnist_simple

# Probably best to keep the non-optional parts of linking hard-coded in the command?
# Maybe good to add $(INCLFLAGS) containing any -I directories in case mlpack is not installed but only built locally?
mnist_simple:
    # If you're not using the Armadillo wrapper, replace `-larmadillo` with linker commands for the
    # BLAS and LAPACK libraries that you are using.
    $(CXX) -o mnist_simple mnist_simple.cpp $(CXXFLAGS) $(LDFLAGS) -lmlpack -larmadillo -lboost_serialization

clean:
    @rm mnist_simple
```

Might be good to put a disclaimer at the top or something, too, that says "you might need to modify this to get it to work on your system, but it should work in most places", or something like that.  It's just an example after all, not meant to build in every possible place. :smile:

Anyway, let me know what you think, these are just my opinions.  We could go with something else if desired.

shrit:
I agree, one makefile by example seems to be a good idea, with a recursive call for the entire project.
I have made it generic because I do not want users to touch the `Target` and `OBJS`. They can change only source files, and add libraries if they want to add new dependencies. I will try to remove some redundant staff and see what we can get.

mnist_simple/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+TARGET := mnist_simple
+SRC := mnist_simple.cpp
+LIBS_NAME := armadillo mlpack boost_serialization
+
+CXX := g++
+CXXFLAGS := -std=c++11 -Wall -Wextra
+LDFLAGS  := -fopenmp
+DBGFLAG := -g
+INCLFLAGS := -I . # Add header directories if there is any.
```
rcurtin, line 15:
```suggestion
INCLFLAGS := -I . # Add header directories if there are any.
```
Maybe a better comment might be `Add header directories for any includes that aren't on the default compiler search path.`?  Up to you.

mnist_simple/Makefile
```diff
@@ -0,0 +1,33 @@
...
+# for the BLAS and LAPACK libraries that you are using.
+
+TARGET := mnist_simple
+SRC := mnist_simple.cpp
+LIBS_NAME := armadillo mlpack boost_serialization
+
+CXX := g++
+CXXFLAGS := -std=c++11 -Wall -Wextra
+LDFLAGS  := -fopenmp
+DBGFLAG := -g
```
rcurtin, line 14:
```suggestion
CXXFLAGS := -std=c++11 -Wall -Wextra -O3 -DNDEBUG
# Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
# CXXFLAGS := -std=c++11 -Wall -Wextra -g -O0
LDFLAGS  := -fopenmp
```
Maybe we should remove this and make it a bit easier to switch back and forth?

Also, it may be useful to add an example `LDFLAGS` that specifies `-L`, in case the user doesn't have mlpack installed or something.  But I'm not sure, I think either way is fine---up to you.

mnist_simple/Makefile
```diff
@@ -0,0 +1,33 @@
...
+INCLFLAGS := -I . # Add header directories if there is any.
+
+OBJS := $(SRC:.c=.o)
```
rcurtin, line 17:
```suggestion
OBJS := $(SRC:.cpp=.o)
```
Not sure, maybe this is needed?  If this is added, then it looks like some other changes are needed so that the `CXXFLAGS` are right for the original compilation.  Anyway, the existing way seems to work; it just skips the compilation to `.o` (which is fine) and compiles directly to the executable.

mnist_simple/Makefile
```diff
@@ -0,0 +1,33 @@
...
+LIBS := $(addprefix -l,$(LIBS_NAME))
+CLEAN_LIST := $(TARGET)
+
+# default rule
+default: all
+
+$(TARGET): $(OBJS)
+	$(CXX) $(CXXFLAGS) $(INCLFLAGS) $(LIBS) $(LDFLAGS) $(OBJS) -o $(TARGET)
```
rcurtin, line 25: To make this work right on my system (I have to specify `-L/home/ryan/src/mlpack/build/lib/` in `LDFLAGS` and `-I/home/ryan/src/mlpack/build/include/ -I/home/ryan/src/mlpack/build/deps/ensmallen-2.11.5/include/` in `INCLFLAGS`), I had to move `$(OBJS) -o $(TARGET)` to the beginning of the command:

```
$(TARGET): $(OBJS)
        $(CXX) $(CXXFLAGS) $(OBJS) -o $(TARGET) $(INCLFLAGS) $(LDFLAGS) $(LIBS)
```

mnist_simple/Makefile
shrit: agreed, the second one is better

shrit:
I think it is ready to be deployed on other examples.

mnist_vae_cnn/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+CXX := g++
+CXXFLAGS := -std=c++11 -Wall -Wextra -O3 -DNDEBUG
+# Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
+# CXXFLAGS := -std=c++11 -Wall -Wextra -g -O0
+LDFLAGS  := -fopenmp
+# LDFLAGS := -L/path/to/mlpack/library/    # if installed locally
+INCLFLAGS := -I . #Add header directories for any includes that aren't on the default compiler search path.
+
+OBJS := $(SRC:.cpp=.o)
```
rcurtin, line 19: Maybe do we need to specify a separate rule for the .cpp -> .o compilation?  The .cpps are being compiled without the `INCLFLAGS` or `CXXFLAGS`:

```
$ head -18 Makefile 
# This is a simple Makefile used to build the example source code.
# This example might requires some modifications in order to work correctly on
# your system.
# If you're not using the Armadillo wrapper, replace `armadillo` with linker commands
# for the BLAS and LAPACK libraries that you are using.

TARGET := mnist_simple
SRC := mnist_simple.cpp
LIBS_NAME := armadillo mlpack boost_serialization

CXX := g++
CXXFLAGS := -std=c++11 -Wall -Wextra -O3 -DNDEBUG
# Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
# CXXFLAGS := -std=c++11 -Wall -Wextra -g -O0
LDFLAGS  := -fopenmp
LDFLAGS := -L/home/ryan/src/mlpack/build/lib/    # if installed locally
INCLFLAGS := -I/home/ryan/src/mlpack/build/include/ -I/home/ryan/src/mlpack/build/deps/ensmallen-2.11.5/include/ #Add header directories for any includes that aren't on the default compiler search path.

$ make
g++ -std=c++11 -Wall -Wextra -O3 -DNDEBUG   -c -o mnist_simple.o mnist_simple.cpp
mnist_simple.cpp:16:10: fatal error: mlpack/core.hpp: No such file or directory
   16 | #include <mlpack/core.hpp>
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [<builtin>: mnist_simple.o] Error 1
```

shrit:
@rcurtin  This one seems to be working, I kept one rule. If it seems good to you, I will deploy these modifications on other examples

rcurtin:
Nice, that seems to work.  It seems also like you could use `+=` instead and `CXXFLAGS` everywhere?  Here's a slightly modified version that might be cleaner:

```
CXX := g++
CXXFLAGS += -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp
# Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
# CXXFLAGS += -std=c++11 -Wall -Wextra -g -O0

LDFLAGS := -fopenmp -L/home/ryan/src/mlpack/build/lib/    # if installed locally
INCLFLAGS := -I/home/ryan/src/mlpack/build/include/ -I/home/ryan/src/mlpack/build/deps/ensmallen-2.11.5/include/ #Add header directories for any includes that aren't on the default compiler search path.

CXXFLAGS += $(INCLFLAGS)
```

(Of course, you would want to include the default `INCLFLAGS` and `LDFLAGS` that you have already.)

I think that using `+=` everywhere could mean that I could set some CXXFLAGS in my environment and have those propagate to the Makefile, which is nice:

```
$ CXXFLAGS="-fail" make
g++ -fail -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp -I/home/ryan/src/mlpack/build/include/ -I/home/ryan/src/mlpack/build/deps/ensmallen-2.11.5/include/    -c -o mnist_simple.o mnist_simple.cpp
g++: error: unrecognized command line option ‘-fail’; did you mean ‘-fnil’?
make: *** [<builtin>: mnist_simple.o] Error 1
```

Anyway, up to you what you want to do.  Overall it looks great to me---want to apply that the changes to all the other Makefiles?  I think this PR is ready once you do that. :+1:

shrit:
@rcurtin I think we can merge it now,

lstm_electricity_consumption/Makefile
```diff
@@ -0,0 +1,36 @@
+# This is a simple Makefile used to build the example source code.
+# This example might requires some modifications in order to work correctly on
```
zoq, line 2: I think we are using the same style guidelines here as for the mlpack repo, so maybe we should adjust the line length here; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping for more details.

lstm_electricity_consumption/Makefile
```diff
@@ -0,0 +1,36 @@
...
+TARGET := lstm_electricity_consumption
+SRC := lstm_electricity_consumption.cpp
+LIBS_NAME := armadillo mlpack boost_serialization
+
+CXX := g++
+CXXFLAGS += -std=c++11 -Wall -Wextra -O3 -DNDEBUG -fopenmp
+# Use these CXXFLAGS instead if you want to compile with debugging symbols and without optimizations.
+# CXXFLAGS += -std=c++11 -Wall -Wextra -g -O0
+LDFLAGS  += -fopenmp
+# LDFLAGS += -L/path/to/mlpack/library/    # if installed locally
```
zoq, line 16:
```suggestion
# LDFLAGS += -L/path/to/mlpack/library/    # If installed locally.
```

Just to be consistent.

lstm_electricity_consumption/Makefile
```diff
@@ -0,0 +1,36 @@
...
+INCLFLAGS := -I . #Add header directories for any includes that aren't on the default compiler search path.
```
zoq, line 17:
```suggestion
# Add header directories for any includes that aren't on the default compiler
# search path.
INCLFLAGS := -I .
```

To be consistent.

lstm_electricity_consumption/Makefile
```diff
@@ -0,0 +1,36 @@
+# This is a simple Makefile used to build the example source code.
```
zoq, line 1: Wondering if we should print a similar message in the build step, not sure the first thing a user will do is to look into the Makefile for more information. Let me know what you think.

lstm_electricity_consumption/Makefile
shrit: Agreed, I need to apply this on all files

lstm_electricity_consumption/Makefile
shrit: I am not sure,  If any system supports posix and have gcc installed, the user should not have any issue with this makefiles. The only possible modifications are setting `LD_FLAGS` and `INCLUDES` if required

shrit:
@rcurtin we can merge now

lstm_electricity_consumption/Makefile
zoq: I guess if there are issues we will see them, so fine with me to go with the current version.

rcurtin:
Thanks @shrit! :+1:

PR #77, opened 2020-04-11
Download mnist dataset from Yann LeCun website.

shrit:
I am trying to create a script that downloads all the datasets used in this repository. However, LSTM electricity consumption and google stock dataset do not seem easy to find online.

The LSTM datasets seem similar to the one from the Kaggle website (maybe part of it) https://www.kaggle.com/uciml/electric-power-consumption-data-set. 
https://www.kaggle.com/uciml/electric-power-consumption-data-set.
Since Kaggle does not allow us to download the dataset unless if we have an account, the script will not be able to download the dataset. 
Is it possible to move a copy to mlpack website? (I do not know the bandwidth limit). We can compress well these data since they are all text, this will reduce the size of the files. 
If this is the case, we can delete all the Kaggle website and allow the user to download the dataset if required.


kartikdutt18:
Great idea!
> Is it possible to move a copy to mlpack website? (I do not know the bandwidth limit). We can compress well these data since they are all text, this will reduce the size of the files.

Yes it is possible, It was discussed on the video meet that we could e-mail @rcurtin the datasets so that they could be uploaded to https://www.mlpack.org/datasets/ . I think more info can be found [here](https://github.com/mlpack/models/pull/3#pullrequestreview-386421401).

download_data_set.sh
```diff
@@ -0,0 +1,26 @@
+echo "This script will download the entire dataset for all the available examples"
```
birm, line 1: Can you add a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) to the top?

birm:
Could you update the readme to reflect the addition of this step?

download_data_set.sh
shrit: Yes, of course. When I asked my self 10 times what is missing at the start :smile:

shrit:
@kartikdutt18 I agree. I remember it from video chat, but I was not able not to find any traces for the website.
Now, I can continue downloading and extracting other datasets.

shrit:
@birm Agreed. I will update the readme as well.

kartikdutt18:
Hey @shrit, Do you think adding arguments to download either all or selected datasets would be better? I think that way user can download only the datasets that he / she requires.

shrit:
@kartikdutt18 I agree, that would be much easier faster and save bandwidth.

download_data_set.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+echo "This script will download the entire dataset for all the available examples"
+echo "Create a dataset directory"
+mkdir -p data
+
+pushd data
+echo "Start downloading the mnist dataset"
+curl http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz --output train-images-idx3-ubyte.gz
```
zoq, line 9: Do you think we should download and extract the archive at the same time? Might be a good idea to save some space.

shrit: @zoq Hmm, What is the possible solution otherwise? The script should delete all the archive at the end after extracting.

zoq: Either write it as a one liner like: `curl url | gunzip -c` or do it in two/three steps: `curl, gunzip, rm` and do that for each file one after another.

shrit: Ok, I see

rcurtin:
Nice, really useful to add this.  But do you think maybe it might be better to implement it in Python so that Windows users have an easier time?  It seems there are a lot of Windows users, so we should probably find some way to be able to serve their needs too. :)  Alternately, we can probably provide a Windows shell script or something, but it seems easier just to use Python libraries to do the same thing.  Let me know what you think. :+1:

shrit:
Ok, Python is a good idea since I do not use windows.
I will follow the same logic by allowing to download dataset separately, but I do not know If I keep the same dataset or update it with a new one? Because in all cases they are not on mlpack website yet.

rcurtin:
It's up to you.  I can store them (in preprocessed form too) on mlpack.org if you like, just let me know what files you want and I'll add them to mlpack.org/datasets/. :+1:

rcurtin:
Ok, I've added the datasets to the website:
 - https://www.mlpack.org/datasets/examples/electricity-usage.csv
 - https://www.mlpack.org/datasets/examples/Google2016-2019.csv
and then we can get the MNIST data directly from the source.

shrit:
@zoq @rcurtin @birm This one is ready for review, no hurry of course.
Thanks.

download_data_set.py
```diff
@@ -0,0 +1,158 @@
...
+def iris_dataset():
+  print("Downloading iris datasets...")
+  iris = requests.get("https://www.mlpack.org/datasets/iris.tar.gz")
+  open("iris.tar.gz", 'wb').write(iris.content)
+  tar = tarfile.open("iris.tar.gz", "r:gz")
+  tar.extractall()
+  tar.close()
+  clean()
+  
+def all_datasets():
```
zoq, line 108: Just tested the script and it works, but I was wondering if it's possible to show a progress bar, because the process takes some time, in fact I opned iftop to see if the script got stuck.

shrit: I agree I thought about that in the first place, it seemed more complicated but I will see what I can do about it. :+1:

zoq: I tested https://stackoverflow.com/a/15645088 the output looks good enough.

shrit: Thanks

shrit:
@zoq tell me what do you think, I think it is ready to get merged

README.md
```diff
@@ -79,3 +79,11 @@ description (just a little bit more than the title):
  - `mnist_vae_cnn`: use a variational autoencoder with convolutional neural
    networks in the encoder and reparametrization networks to recognize the MNIST
    digits
+   
+### 6. Datasets
+
+All the required dataset needed by the examples can be downloaded using the
+provided script. You will have to execute `download_dataset.sh` and it will
```
rcurtin, line 86:
```suggestion
provided script. You will have to execute `download_dataset.py` and it will
```
Also I guess either we need to change the filename to `download_dataset.py` (that would be my slight preference, actually `download_datasets.py` might be more accurate :)), or update this documentation to `download_data_set.py`. :+1:

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+import sys
+import tarfile
+import textwrap
+from tqdm import tqdm
+import requests
+import shutil
+
+''' This function is originally written by Joseph Redmon (pjreddie).
+    The original source code can be found here:
+      https://pjreddie.com/projects/mnist-in-csv/
```
rcurtin, line 17: Technically there is no license with his code.  I know it's super pedantic to bring this up, but should we get in touch and ask if it's okay if we incorporate the code here?

Also I love his CV: https://pjreddie.com/static/Redmon%20Resume.pdf

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+      t.update(len(data))
+      f.write(data)
+  t.close()
+  if total_size != 0 and t.n != total_size:
+    print("ERROR, something went wrong")
+
+def mnist_dataset():
+  print("Start downloading the mnist dataset")
+  train_features = requests.get(
+  "http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz")
```
rcurtin, line 86:
```suggestion
      "http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz")
```

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+  progress_bar("train_features.gz", train_features)
+  ungzip("train_features.gz", "train_features.ubytes")
+  
+  train_labels = requests.get(
+  "http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz")
```
rcurtin, line 91:
```suggestion
      "http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz")
```

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+  progress_bar("train_labels.gz", train_labels)
+  ungzip("train_labels.gz", "train_labels.ubytes")
+
+  test_features = requests.get(
+  "http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz")
```
rcurtin, line 96:
```suggestion
      "http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz")
```

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+  progress_bar("test_features.gz", test_features)
+  ungzip("test_features.gz", "test_features.ubytes")
+  
+  test_labels = requests.get(
+  "http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz")
```
rcurtin, line 101:
```suggestion
      "http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz")
```

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+  progress_bar("test_labels.gz", test_labels)
+  ungzip("test_labels.gz", "test_labels.ubytes")
+
+  print("Converting mnist ubytes images files into csv...")
+  print("This might take a while...")
+  print("Converting features images...")
+  convert("train_features.ubytes", "train_labels.ubytes",
+  "mnist_train.csv", 60000)
```
rcurtin, line 109:
```suggestion
  convert("train_features.ubytes", "train_labels.ubytes", "mnist_train.csv", 60000)
```

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+  print("Converting label images...")
+  convert("test_features.ubytes", "test_labels.ubytes",
+  "mnist_test.csv", 10000)
```
rcurtin, line 112:
```suggestion
  convert("test_features.ubytes", "test_labels.ubytes", "mnist_test.csv", 10000)
```

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+  stock_exchange_dataset()
+  iris_dataset()
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(
+        prog='Download dataset script',
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description=textwrap.dedent('''\
+        Dataset script!
```
rcurtin, line 146:
```suggestion
        Dataset downloading script!
```
Just a suggestion, up to you what you'd like to do. :)

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+        mnist : will download mnist dataset
+        electricity : will download electricty_consumption_dataset
+        stock : will download stock_exchange dataset
+        all : will download all datasets for all examples
+        '''))
+ 
+    parser.add_argument('--dataset_name', metavar="dataset name", type=str, help="Enter dataset name to download")
+    args = parser.parse_args()
+
+    if len(sys.argv)==1:
```
rcurtin, line 164:
```suggestion
    if len(sys.argv) == 1:
```
Sorry to be a bit pedantic. :)

download_data_set.py
```diff
@@ -0,0 +1,184 @@
...
+      parser.print_help(sys.stderr)
+      sys.exit(1)
```
rcurtin, line 166: I dunno, maybe we should just assume the default of `all` in this case?

shrit: Agreed

download_data_set.py
shrit: @rcurtin Yes, I had a similar thought of sending an email asking him. The function looks good, so I copied technically (With only progress bar added).
Would you like me to send him an email? or you can handle it.

True, I love his CV too, it is kind of CV you will never forget.

shrit:
@zoq I have moved the script into a `tools` directory

README.md
```diff
@@ -79,3 +79,10 @@ description (just a little bit more than the title):
  - `mnist_vae_cnn`: use a variational autoencoder with convolutional neural
    networks in the encoder and reparametrization networks to recognize the MNIST
    digits
+   
+### 6. Datasets
+
+All the required dataset needed by the examples can be downloaded using the
+provided script in the `tools` direcotory. You will have to execute `download_dataset.py`
```
shrit, line 86:
```suggestion
provided script in the `tools` directory. You will have to execute `download_dataset.py`
```

download_data_set.py
rcurtin: Ah, sorry, I forgot to actually do this.  Let me do it now and CC you. :+1:

rcurtin: Once I get a response, I'll post about it here.

shrit: @rcurtin Thanks :+1:

rcurtin: Quick response:

```
Date: Mon, 18 May 2020 16:23:51 -0700                                                                                                                                          
From: Joseph Redmon <pjreddie@cs.washington.edu>                                                                                                                               
To: Ryan Curtin <ryan@ratml.org>                                                                                                                                               
Cc: pjreddie@uw.edu, shrit@outlook.com                                                                                                                                         
Subject: Re: License for MNIST CSV loading code                                                                                                                                
X-Spam-Level:                                                                                                                                                                  
X-Mailer: iPhone Mail (17D50)                                                                                                                                                  
                                                                                                                                                                               
You can use it consider it public domain   
```

So, good to go. :+1:

PR #72, opened 2020-03-29
Add LSTM example using Notebooks.

kartikdutt18:
I am still getting used to Xeus. Just one thing that I need to fix is that model needs to train for more than one epoch. Other than that let me know what you think. Should comments and headings be more elaborate. Thanks a lot.

PR #83, opened 2020-05-07
Simplify Optimizers used in examples.

kartikdutt18:
This is in reference to a comment in #72 by @rcurtin to simplify optimizers. Here instead of using  `SGD<AdamUpdate>` with AdamUpate we can simply use `Adam`

Let me know what you think. Thanks a lot.

### DIFF
--- a/lstm_electricity_consumption/lstm_electricity_consumption.cpp
+++ b/lstm_electricity_consumption/lstm_electricity_consumption.cpp
@@ -240,14 +240,16 @@ int main()
       model.Add<Linear<> >(H1, outputSize);
     }
 
-    // Set parameters for the Stochastic Gradient Descent (SGD) optimizer.
-    SGD<AdamUpdate> optimizer(
-        STEP_SIZE, // Step size of the optimizer.
-        BATCH_SIZE, // Batch size. Number of data points that are used in each iteration.
-        trainData.n_cols * EPOCHS, // Max number of iterations.
-        1e-8, // Tolerance.
-        true, // Shuffle.
-        AdamUpdate(1e-8, 0.9, 0.999)); // Adam update policy.
+    // Set parameters for the Adam optimizer.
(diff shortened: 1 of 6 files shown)


lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -240,14 +240,16 @@ int main()
...
+    // Set parameters for the Adam optimizer.
+    ens::Adam optimizer(
+    STEP_SIZE, // Step size of the optimizer.
+    BATCH_SIZE, // Batch size. Number of data points that are used in each iteration.
+    0.9, // Exponential decay rate for the first moment estimates.
+    0.999, // Exponential decay rate for the weighted infinity norm estimates.
+    1e-8, // Value used to initialise the mean squared gradient parameter.
+    trainData.n_cols * EPOCHS, // Max number of iterations.
+    1e-8, // Tolerance.
+    true);
```
zoq, line 252:
```suggestion
        STEP_SIZE, // Step size of the optimizer.
        BATCH_SIZE, // Batch size. Number of data points that are used in each iteration.
        0.9, // Exponential decay rate for the first moment estimates.
        0.999, // Exponential decay rate for the weighted infinity norm estimates.
        1e-8, // Value used to initialise the mean squared gradient parameter.
        trainData.n_cols * EPOCHS, // Max number of iterations.
        1e-8, // Tolerance.
        true);
```

To be consistent with the rest of the codebase.

kartikdutt18: Ahh, Sorry I missed that one. Thanks a lot for the review.

saksham189:
Thanks for another great contribution!

kartikdutt18:
Thanks @saksham189, @zoq for reviews.
Regards.

PR #86, opened 2020-05-08
Adding the same license as mlpack

shrit:


shrit:
I have noticed that there is no license in this repository, but I am not sure is we need one.

shrit:
Perfect, I will add one in mlpack models

PR #87, opened 2020-05-10
Use this repo's start year

birm:
Following #86 and https://github.com/mlpack/models/pull/6, change start date in copyright notice to 2017


shrit:
Thanks,

PR #90, opened 2020-05-19
Added cartpole-dqn notebook

nishantkr18:


q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Set up the network.\n",
+    "  SimpleDQN<> model(4, 256, 128, 2);"
```
kartikdutt18, line 93: Hey @nishantkr18, Could you please fix the spacing here. Thanks.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+      "action:    1.0000\n",
+      " Current step: 3 current reward: 3\n",
+      "action:         0\n",
+      " Current step: 4 current reward: 4\n",
+      "action:    1.0000\n",
+      " Current step: 5 current reward: 5\n",
+      "action:    1.0000\n",
+      " Current step: 6 current reward: 6\n",
+      "action:         0\n",
+      " Current step: 7 current reward: 7\n",
```
kartikdutt18, line 322: Hey @nishantkr18, Any chance that we reduce the number of steps here and we can add a comment that increasing them would lead to better results.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// You can easily run this notebook at https://lab.mlpack.org/"
```
zoq, line 9: I used a badge in #74, maybe we should do the same here as well: `"[mlpack-lab Image](https://lab.mlpack.org)"`.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+    "#include <mlpack/methods/ann/layer/layer.hpp>\n",
+    "#include <mlpack/methods/ann/loss_functions/mean_squared_error.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/q_learning.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/environment/mountain_car.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/environment/acrobot.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/environment/cart_pole.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/environment/double_pole_cart.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp>\n",
+    "#include <mlpack/methods/reinforcement_learning/training_config.hpp>"
```
zoq, line 38: Some of the includes are not needed, we should remove the ones that are not needed to keep it simple.

q_learning/cartpole_dqn.ipynb
zoq, line 93: Let's remove the extra whitespace here.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+    "    averageReturn(episodeReturn);\n",
+    "    episodes += 1;\n",
+    "    \n",
+    "    std::cout << \"Average return: \" << averageReturn.mean()\n",
+    "        << \" Episode return: \" << episodeReturn << std::endl;\n",
+    "    \n",
+    "    if (episodes > 1000)\n",
+    "    {\n",
+    "      std::cout << \"Cart Pole with DQN failed.\" << std::endl;\n",
+    "      converged = false;\n",
```
zoq, line 205: The parameter was used for testing, in this case I guess people might wander why we use it here.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+    "}"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 13,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#include <gym/environment.hpp>"
```
zoq, line 229: Let's put this to the top, also add a comment to mention this is used to display the video.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#include \"xwidgets/ximage.hpp\"\n",
```
zoq, line 238: See comment above.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+    "size_t episodes = 0;\n",
+    "bool converged = true;\n",
+    "\n",
+    "while (true)\n",
+    "{\n",
+    "    double episodeReturn = agent.Episode();\n",
+    "    averageReturn(episodeReturn);\n",
+    "    episodes += 1;\n",
+    "    \n",
+    "    std::cout << \"Average return: \" << averageReturn.mean()\n",
```
zoq, line 199: Maybe we should only print every tenth, to keep the output simple.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+     "output_type": "stream",
+     "text": [
+      "https://gym.kurg.org/647380835f904/output.webm\n"
+     ]
+    }
+   ],
+   "source": [
+    "env.close();\n",
+    "std::string url = env.url();\n",
+    "std::cout << url << std::endl;"
```
zoq, line 755: No need to print the url here.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+       "A Jupyter widget"
+      ]
+     },
+     "execution_count": 22,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "auto video = xw::video_from_url(url).finalize();\n",
```
zoq, line 780: Might be interesting to show the agent at different stages.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// You can easily run this notebook at https://lab.mlpack.org/"
+   ]
+  },
+  {
+   "cell_type": "code",
```
zoq, line 13: Would be great if you can add a simple description what the notebook is all about, some lines regarding the method and the env, and what the goal is.

zoq:
>     3\. Could you also please add comments to make this like a tutorial since the purpose of this repo is to get beginners started with mlpack.
>         An example can be seen [here](https://github.com/mlpack/examples/blob/master/neural_network_regression/neural_network_regression.ipynb) and [here](https://github.com/mlpack/examples/blob/master/lstm_stock_prediction/lstm_multivariate_time_series_prediction.ipynb). Let me know what you think, thanks a lot for this amazing contribution. It's nice to see new projects being added to this repo.
>         Regards.

Personally I would keep it simple, so a introduction is probably good enough here, and use code comments to point out important parts.

kartikdutt18:
> Personally I would keep it simple, so a introduction is probably good enough here, and use code comments to point out important parts.

Sure, that makes sense. I guess we would still add a cpp file though right?

zoq:
Yes, good idea.

kartikdutt18:
> Yes, good idea.

Great. Thanks a lot.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,801 @@
...
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Average return: 25 Episode return: 25\n",
```
favre49, line 143: Could we also format the output here better? It's pretty difficult to read.

q_learning/cartpole_dqn.ipynb
nishantkr18: By stages, do u mean in between training? If that's so, then we'd have to train for a while, pause and test the agent, generate the video, then again continue training, and so on..

q_learning/cartpole_dqn.ipynb
nishantkr18: How can we decrease the steps here? These are the number of  steps taken by the testing agent.. So it depends on the agent. Please correct me if I am wrong :)

q_learning/cartpole_dqn.ipynb
nishantkr18: Is that ok now?

q_learning/cartpole_dqn.ipynb
nishantkr18: Done :+1:

nishantkr18:
> Sure, that makes sense. I guess we would still add a cpp file though right?

We will be implementing more extensions for dqn, and probably be adding them here.. which would mean for every new extension, we'd have to add a cpp file along with it. And any changes in the implementation would mean making the same changes in 2 places (cpp and notebook). Should that be the case?

q_learning/cartpole_dqn.ipynb
zoq: I think @kartikdutt18 only meant to reduce the output.

q_learning/cartpole_dqn.ipynb
zoq: Two stages would be just fine, what I had in mind is break after like like `average return: is around 21` create a video and continue the training of the same agent afterwards and create another video. Does that sound reasonable?

"In that case, I think train, test functions need to be created right?"

Not sure I understand, what you mean.

q_learning/cartpole_dqn.ipynb
kartikdutt18: Right, Sorry I wasn't clear earlier. Maybe we could print output every 10 steps or something like that. It's better to keep the notebook short.
Thanks a lot.

kartikdutt18:
> Should that be the case?

Yes, So if we have a different notebook for each extention then we should also have a cpp file for it. If you make the changes in one notebook then those changes should also be reflected in the corresponding cpp file.
Regards.

zoq:
Btw. @jeffin143 suggested to test https://www.reviewnb.com/, to make the review process for notebook easier, I installed the app, you should be able to see it here: https://app.reviewnb.com/mlpack/examples/

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,777 @@
+{
```
zoq, line 1: <p>I guess it makes sense to minify the output here as well, what do you think?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/90/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

q_learning/cartpole_dqn.ipynb
nishantkr18: > Two stages would be just fine, what I had in mind is break after like like average return: is around 21 create a video and continue the training of the same agent afterwards and create another video. Does that sound reasonable?

Yeah, sounds good..

> Not sure I understand, what you mean.

I meant, wrapping the testing and training parts of the code inside separate functions, something like, 
```
// Train till avg rewards reach 21
train(21)

// Test the agent and get video
test()

// Continue training till avg rewards 50
train(50)

// test and get video
test()
```

zoq: Not sure that's needed, but if you think that would improve the overall readability, sure go for it. Btw. thanks for keeping up with all the comments 👍

nishantkr18: > Not sure that's needed.

I think I am missing something. If we dont do that, we'd have to copy-paste the training and testing block twice right?

kartikdutt18:
<p>This extention is amazing!</p>

jeffin143:
you are fast @zoq , I guess there is integration or something that will host it for every pr and give us a button to redirect to exact pr notebook, I will check that out  and let you know

q_learning/cartpole_dqn.ipynb
favre49: Yup :)

kartikdutt18:
They also support testing for notebooks as well. Here is their [blog](https://blog.reviewnb.com/reproducible-notebooks/).

q_learning/cartpole_dqn.ipynb
nishantkr18: I could'nt get the testing block inside a function.. so had to copy-paste it.. I hope there's a better way

nishantkr18:
> Yes, So if we have a different notebook for each extention then we should also have a cpp file for it. If you make the changes in one notebook then those changes should also be reflected in the corresponding cpp file.
> Regards.

Alright, I'll do the changes once we are finalized with the notebook. Thanks :+1:

q_learning/cartpole_dqn.ipynb
nishantkr18: will the current change work? @kartikdutt18 sorry I didn't read your reply before making the change..

kartikdutt18: No worries. Yep, Looks good. Thanks a lot for making the changes.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "#include <mlpack/methods/reinforcement_learning/training_config.hpp>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Used to run the agent on gym's environment (provided externally) for testing\n",
```
zoq, line 54:
```suggestion
    "// Used to run the agent on gym's environment (provided externally) for testing.\n",
```

Pedantic style issue.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[![mlpack-lab Image](https://img.shields.io/endpoint?url=https%3A%2F%2Flab.kurg.org%2Fstatus%2Fstatus.json)](https://lab.mlpack.org)\n",
+    "\n",
+    "You can easily run this notebook at https://lab.mlpack.org/\n",
+    "\n",
+    "This notebook shows how to get started with training reinforcement learning agents, particularly DQN agents, using mlpack. Here, we train a [Simple DQN](https://www.cs.toronto.edu/~vmnih/docs/dqn.pdf) agent to get high scores for the [CartPole](https://gym.openai.com/envs/CartPole-v0) environment. \n",
+    "\n",
+    "Mlpack contains non-gui implementations of some of OpenAI gym's environments. In this notebook, we use one such environment for training the agent, as it is fast to train on.\n",
```
zoq, line 13:
```suggestion
    "mlpack contains non-gui implementations of some of OpenAI gym's environments. In this notebook, we use one such environment for training the agent, as it is fast to train on.\n",
```

I try to use mlpack everwhere, but I have seen all sorts of variations, MLPACK, mlpack, MLpack, MLPack.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "size_t episodes = 0;"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Function to train the agent on mlpack's own implementation of the CartPole environment:\n",
```
zoq, line 186:
```suggestion
    "// Function to train the agent on mlpack's own implementation of the CartPole environment.\n",
```

Another pedantic style issue.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "  while (true)\n",
+    "    {\n",
+    "        double episodeReturn = agent.Episode();\n",
+    "        averageReturn(episodeReturn);\n",
+    "        episodes += 1;\n",
+    "\n",
+    "        if(episodes % 10 == 0)\n",
+    "            std::cout << \"Average return: \" << averageReturn.mean()\n",
+    "                << \"\\t Episode return: \" << episodeReturn \n",
+    "                << \"\\t Episode number: \" << episodes << std::endl;\n",
```
zoq, line 199: Do you mind to use `{}` here, I think it would improve the readability.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Function to train the agent on mlpack's own implementation of the CartPole environment:\n",
+    "void train(int threshold)\n",
```
zoq, line 187: I guess since we use `size_t` everwhere it makes sense to use it here as well, also we might want to use `const size_t threshold`.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "\n",
+    "// Creating and setting up the gym environment for testing.\n",
+    "gym::Environment env(\"gym.kurg.org\", \"4040\", \"CartPole-v0\");\n",
+    "env.monitor.start(\"./dummy/\", true, true);\n",
+    "\n",
+    "// Resets the environment.\n",
+    "env.reset();\n",
+    "env.render();\n",
+    "\n",
+    "// Initializing variables\n",
```
zoq, line 294:
```suggestion
    "// Initializing variables.\n",
```

See https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments for more information.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "double totalReward = 0;\n",
+    "size_t totalSteps = 0;\n",
+    "\n",
+    "// Testing the agent on gym's environment\n",
```
zoq, line 298:
```suggestion
    "// Testing the agent on gym's environment.\n",
```

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "while (1)\n",
+    "  {\n",
+    "    // State from the environment is passed to the agent's internal representation\n",
```
zoq, line 301:
```suggestion
    "    // State from the environment is passed to the agent's internal representation.\n",
```

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "    agent.State().Data() = env.observation;\n",
+    "\n",
+    "    // with the given state, the agent performs an action according to its defined policy\n",
```
zoq, line 304:
```suggestion
    "    // With the given state, the agent performs an action according to its defined policy.\n",
```

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "    totalSteps += 1;\n",
+    "\n",
+    "    if (env.done)\n",
+    "    {\n",
+    "      std::cout << \" Total steps: \" << totalSteps << \"\\t Total reward: \"\n",
+    "        << totalReward << std::endl;\n",
+    "      break;\n",
+    "    }\n",
+    "    \n",
+    "    // Uncomment the following lines to see the reward and action in each step\n",
```
zoq, line 321:
```suggestion
    "    // Uncomment the following lines to see the reward and action in each step.\n",
```

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,462 @@
...
+    "env.reset();\n",
+    "env.render();\n",
+    "\n",
+    "// Initializing variables\n",
+    "double totalReward = 0;\n",
+    "size_t totalSteps = 0;\n",
+    "\n",
+    "// Testing the agent on gym's environment\n",
+    "while (1)\n",
+    "  {\n",
```
zoq, line 412: Looks like the indentation is off, same for the second while loop.

zoq:
Looking at the output, I wonder why it seems to stuck at an average return of 100, it should be much higher.

zoq:
I tested `PrioritizedReplay` as well, which works better (180 reward), but that's still not the 190 we need. Not sure what the reason is, maybe our env implementation is not the same as gym, maybe the network is too small, or there is something wrong the the QLearning implementation.

nishantkr18:
> I tested `PrioritizedReplay` as well, which works better (180 reward), but that's still not the 190 we need. Not sure what the reason is, maybe our env implementation is not the same as gym, maybe the network is too small, or there is something wrong the the QLearning implementation.

I think we'd have to tune the hyperparameters a bit maybe, like increasing the batch size, using a bigger/smaller network. I have a pytorch implemetation to compare the results with.

q_learning/cartpole_dqn.ipynb
nishantkr18: I think I should keep 4 spaces as indentation (cannot change it on lab.mlpack.org). So i've made changes everywhere. I hope thats fine

q_learning/cartpole_dqn.ipynb
nishantkr18: Yeah, I better remove that. "overly verbose" :laughing:

zoq:
Maybe, we should also test other network weight initialization strategies. Would make sense to show a model that is able to solve the env, so around 190 reward over multiple runs.

nishantkr18:
> Maybe, we should also test other network weight initialization strategies. Would make sense to show a model that is able to solve the env, so around 190 reward over multiple runs.

Yeah, I think people mostly use xavier weight initialization.. But I think it matters for deep networks(maybe for >5 layers). As far as I remember, gaussionInit is not used for deep networks as the activations diminish to zero in deep networks. Please correct me if im wrong

zoq:
Agreed, easy to test different strategies.

q_learning/cartpole_dqn.ipynb
zoq: Usually we use 2 spaces.

nishantkr18: Alright, I'll make the changes.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,444 @@
...
+   "source": [
+    "[![mlpack-lab Image](https://img.shields.io/endpoint?url=https%3A%2F%2Flab.kurg.org%2Fstatus%2Fstatus.json)](https://lab.mlpack.org)\n",
+    "\n",
+    "You can easily run this notebook at https://lab.mlpack.org/\n",
+    "\n",
+    "This notebook shows how to get started with training reinforcement learning agents, particularly DQN agents, using mlpack. Here, we train a [Simple DQN](https://www.cs.toronto.edu/~vmnih/docs/dqn.pdf) agent to get high scores for the [CartPole](https://gym.openai.com/envs/CartPole-v0) environment. \n",
+    "\n",
+    "mlpack contains non-gui implementations of some of OpenAI gym's environments. In this notebook, we use one such environment for training the agent, as it is fast to train on.\n",
+    "\n",
+    "As for testing, we make the agent run on OpenAI Gym toolkit's GUI interface provided through a distributed infrastructure (TCP API). A video of the trained agent can be seen in the end."
```
favre49, line 15: Should we link to the gym-tcp-api here?

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,444 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[![mlpack-lab Image](https://img.shields.io/endpoint?url=https%3A%2F%2Flab.kurg.org%2Fstatus%2Fstatus.json)](https://lab.mlpack.org)\n",
+    "\n",
+    "You can easily run this notebook at https://lab.mlpack.org/\n",
+    "\n",
+    "This notebook shows how to get started with training reinforcement learning agents, particularly DQN agents, using mlpack. Here, we train a [Simple DQN](https://www.cs.toronto.edu/~vmnih/docs/dqn.pdf) agent to get high scores for the [CartPole](https://gym.openai.com/envs/CartPole-v0) environment. \n",
+    "\n",
+    "mlpack contains non-gui implementations of some of OpenAI gym's environments. In this notebook, we use one such environment for training the agent, as it is fast to train on.\n",
```
favre49, line 13:
```suggestion
    "mlpack contains non-GUI implementations of some of OpenAI gym's environments. In this notebook, we use one such environment for training the agent, as it is fast to train on.\n",
```

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,444 @@
...
+    "        }\n",
+    "    }\n",
+    "}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Let the training begin:"
```
favre49, line 228:
```suggestion
    "## Let the training begin"
```

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,444 @@
...
+       "A Jupyter widget"
+      ]
+     },
+     "execution_count": 16,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "// TESTING BLOCK\n",
```
favre49, line 289: I think we can remove this?

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,444 @@
...
+    "\n",
+    "auto video = xw::video_from_url(url).finalize();\n",
+    "video"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## A little more training.."
```
favre49, line 341:
```suggestion
    "## A little more training..."
```

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,442 @@
...
+      "Average return: 40.66\t Episode return: 161\t Episode number: 50\n",
+      "Average return: 57.25\t Episode return: 141\t Episode number: 60\n",
+      "Average return: 73.2\t Episode return: 200\t Episode number: 70\n",
+      "Average return: 74.125\t Episode return: 105\t Episode number: 80\n",
+      "Average return in deterministic test: 157.7\n"
+     ]
+    }
+   ],
+   "source": [
+    "// Training the same agent till average return reaches 60\n",
```
favre49, line 362:
```suggestion
    "// Training the same agent till average return reaches 75\n",
```

zoq: We have to update the number if we have a working model, because:

"CartPole-v0 defines "solving" as getting average reward of 195.0 over 100 consecutive trials."

nishantkr18: I tried keeping the number high, the problem is like u said, it never goes beyond an avg of 100. So, I'm now trying to set proper hyperparams first.. If taht does't work, we need to check the implementations..

nishantkr18: > "CartPole-v0 defines "solving" as getting average reward of 195.0 over 100 consecutive trials."

Since we are calculating average as a running_stat from the begining, I think it might take the `total Average` a long time to come close to `195`. But I think Its able to achieve a consecutive 100 trial avg score of >195.

zoq: Yes, I think it's reasonable to keep the number around 100, I guess if we have a model that works we can also serialize/save it and add a comment about using a pre-trained model.

kartikdutt18:
Hi @favre49, I think the notebook is complete but we should also try to add the cpp file as well. I think that suits the general structure of repo i.e. to maintain a notebook and cpp file. Let me know what you think.

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,498 @@
...
+      "Average return: 22.44\t Episode return: 43\t Episode number: 100\n",
+      "Average return: 26.1545\t Episode return: 52\t Episode number: 110\n",
+      "Average return: 33.9\t Episode return: 110\t Episode number: 120\n",
+      "Average return in 100 consecutive trials : 160.13\n"
+     ]
+    }
+   ],
+   "source": [
+    "// Training the agent till the total average return reaches 35.\n",
+    "train(35)"
```
zoq, line 260: We might want to use a lower number here, because the output is already quite good and the idea was to show that the agent is able to solve this over time.

nishantkr18: Hmm.. Actually I first kept 25.. It seemed the model performed very badly.. So thought to maybe increase the threshold(to show partial training).. I'm not sure which threshold should serve as the sweet spot :)

favre49:
@kartikdutt18 You're right, my bad. I knew something was missing, just couldn't put my finger on it :) Thanks for pointing it out

q_learning/cartpole_dqn.ipynb
zoq: The current solution looks fine for me.

zoq:
@nishantkr18 @favre49 @kartikdutt18 since this comes with an extra dependency (https://github.com/zoq/gym_tcp_api) that we have to incorporate into the build process, I would open a another PR for the cpp file.

favre49:
One thing before we merge. In [this](https://github.com/mlpack/examples/pull/74#issuecomment-615995928) comment, Ryan suggested we have one directory for each example notebook. I guess we should maintain a standard across the repo. Shall we change the name of the folder?

P.S. Sorry for commenting on the wrong issue before this

zoq:
Hm, I guess `cartpole_dqn` might be a good name for the folder, we use the same name for the notebook, but I think that's fine.

nishantkr18:
> Hm, I guess `cartpole_dqn` might be a good name for the folder, we use the same name for the notebook, but I think that's fine.

Yeah, for now I guess we can keep it that way.. But I was thinking of adding other extensions(noisy, dueling etc) in this folder itself. Will that be ok? or should each extension have a separate folder?

q_learning/cartpole_dqn.ipynb
```diff
@@ -0,0 +1,501 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Set up the network.\n",
+    "SimpleDQN<> model(4, 128, 32, 2);"
```
favre49, line 122: If I got the API right, I think this line is the only line that would change across notebooks? Perhaps it would be overkill to create multiple folders for this.

zoq:
Let's merge this as is and if needed we will change the folder or whatever is needed.

PR #91, opened 2020-05-20
Use ratml instead of masterblaster to configure armadillo.

kartikdutt18:
This should fix travis.ci builds.


birm:
It didn't occur to me to pronounce ratml as rattle until now, but I like it.

kartikdutt18:
> It didn't occur to me to pronounce ratml as rattle until now, but I like it.

xD. Let me fix that.

kartikdutt18:
Thanks @birm, @favre49 for the reviews.

PR #93, opened 2020-06-03
Update lstm_electricity_consumption.cpp

pandeynandancse:


birm:
Thanks and welcome to mlpack! I would recommend cherry picking or merging the change in #94 into this as well.

zoq:
> Thanks and welcome to mlpack! I would recommend cherry picking or merging the change in #94 into this as well.

Looks like this includes the change from #94.

birm:
#94 only touches lstm_stock_prediction.cpp which is not modified here.

PR #95, opened 2020-06-07
Update README

shrit:


shrit:
I think we can update readme at this level

PR #98, opened 2020-07-14
Early stop

shrit:
This pull request updates the early stop callback to stop the training when the loss stops decreasing on the validation loss rather than the training loss.

### DIFF
--- a/mnist_batch_norm/mnist_batch_norm.cpp
+++ b/mnist_batch_norm/mnist_batch_norm.cpp
@@ -148,10 +148,10 @@ int main()
                   [&](const arma::mat& /* param */)
                   {
                     double validationLoss =
-                    model.Evaluate(validX, validY);
-                    std::cout << "Validation loss: "
-                              << validationLoss
-                              << std::endl;
+                        model.Evaluate(validX, validY);
+                        std::cout << "Validation loss: "
+                                  << validationLoss
+                                  << std::endl;
                     return validationLoss;
                   }));
   mat predOut;
--- a/mnist_batch_norm/mnist_batch_norm.cpp
+++ b/mnist_batch_norm/mnist_batch_norm.cpp
@@ -147,11 +147,9 @@ int main()
               ens::EarlyStopAtMinLoss<arma::mat>(
                   [&](const arma::mat& /* param */)
                   {
-                    double validationLoss =
-                        model.Evaluate(validX, validY);
-                        std::cout << "Validation loss: "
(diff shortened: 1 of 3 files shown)


mnist_batch_norm/mnist_batch_norm.cpp
```diff
@@ -144,8 +144,16 @@ int main()
...
-              ens::EarlyStopAtMinLoss());
-
+              ens::EarlyStopAtMinLoss<arma::mat>(
+                  [&](const arma::mat& /* param */)
+                  {
+                    double validationLoss =
+                    model.Evaluate(validX, validY);
+                    std::cout << "Validation loss: "
+                              << validationLoss
+                              << std::endl;
```
zoq, line 154:
```suggestion
                        model.Evaluate(validX, validY);
                        std::cout << "Validation loss: "
                                  << validationLoss
                                  << std::endl;
```

Hopefully the idendation is correct (tab twice - 4 spaces).

mnist_cnn/mnist_cnn.cpp
```diff
@@ -179,7 +179,16 @@ int main()
               ens::PrintLoss(),
               ens::ProgressBar(),
               // Stop the training using Early Stop at min loss.
-              ens::EarlyStopAtMinLoss());
+              ens::EarlyStopAtMinLoss<arma::mat>(
+                  [&](const arma::mat& /* param */)
+                  {
+                    double validationLoss =
+                    model.Evaluate(validX, validY);
```
zoq, line 186: See comment above.

mnist_simple/mnist_simple.cpp
```diff
@@ -130,7 +130,16 @@ int main()
               ens::PrintLoss(),
               ens::ProgressBar(),
               // Stop the training using Early Stop at min loss.
-              ens::EarlyStopAtMinLoss());
+              ens::EarlyStopAtMinLoss<arma::mat>(
+                  [&](const arma::mat& /* param */)
+                  {
+                    double validationLoss =
+                    model.Evaluate(validX, validY);
```
zoq, line 137: See comment above.

mnist_batch_norm/mnist_batch_norm.cpp
shrit: I think the 4 spaces are special lambda right?

shrit: I see because I have just cut the lines

PR #100, opened 2020-07-19
Adding notebook for solving acrobot using dqn

nishantkr18:


q_learning/acrobot_dqn.ipynb
```diff
@@ -0,0 +1,567 @@
...
+    "  agent.State().Data() = envTest.observation;\n",
+    "\n",
+    "  // With the given state, the agent selects an action according to its defined policy.\n",
+    "  agent.SelectAction();\n",
+    "\n",
+    "  // Action to take, decided by the policy.\n",
+    "  arma::mat action = {double(agent.Action().action)};\n",
+    "\n",
+    "  envTest.step(action);\n",
+    "  totalReward += env.reward;\n",
```
zoq, line 528:
```suggestion
    "  totalReward += envTest.reward;\n",
```

That should fix the reward issue.

zoq:
Currently testing with:

```
GreedyPolicy<DiscreteActionEnv> policy(1.0, 1000, 0.1, 0.99);
RandomReplay<DiscreteActionEnv> replayMethod(32, 2000);
```

and I got https://gym.kurg.org/630f0a4878ef4/output.webm as a result.

q_learning/acrobot_dqn.ipynb
```diff
@@ -0,0 +1,567 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[![Binder](https://mybinder.org/badge_logo.svg)](https://lab.mlpack.org/v2/gh/mlpack/examples/master?urlpath=lab%2Ftree%2Fq_learning%2Fcartpole_dqn.ipynb)\n",
```
favre49, line 7: We should change this path to acrobot

q_learning/acrobot_dqn.ipynb
nishantkr18: Oh, sorry about that..

q_learning/acrobot_dqn.ipynb
nishantkr18: Oh yes! missed it.

zoq:
> Currently testing with:
> 
> ```
> GreedyPolicy<DiscreteActionEnv> policy(1.0, 1000, 0.1, 0.99);
> RandomReplay<DiscreteActionEnv> replayMethod(32, 2000);
> ```
> 
> and I got https://gym.kurg.org/630f0a4878ef4/output.webm as a result.

@nishantkr18 have you seen good results with Prioritized Replay?

nishantkr18:
> @nishantkr18 have you seen good results with Prioritized Replay?

Well, I felt no differnce. Prioritized also gave a similar result. But maybe it was more sample efficient. Why do u ask btw?

rcurtin:
I didn't review this thoroughly but I took a quick look, and this is really cool!  mlpack notebooks are awesome. :100:

zoq:
> > @nishantkr18 have you seen good results with Prioritized Replay?
> 
> Well, I felt no differnce. Prioritized also gave a similar result. But maybe it was more sample efficient. Why do u ask btw?

I'm not sure if that is a problem with the replay strategy or not, since when I switch to `RandomReplay` the result I got was the video above.

zoq:
Did some further testing looks like the result is better if you increase the model size and use `RandomReplay` but it's not consistently, at least I couldn't produce a model to worked in each run. Do you think we should test some other env and see if the agent is able to solve it?

nishantkr18:
> Did some further testing looks like the result is better if you increase the model size and use `RandomReplay` but it's not consistently, at least I couldn't produce a model to worked in each run. Do you think we should test some other env and see if the agent is able to solve it?

Yeah, let me go ahead and do that. Btw, did you run the tests online on notebook or locally?

nishantkr18:
```c++
/**
 * @file example.cpp
 * @author Marcus Edel
 *
 * Simple random agent.
 */
#include <mlpack/core.hpp>

#include "../../mlpack/src/mlpack/methods/ann/ffn.hpp"
#include "../../mlpack/src/mlpack/methods/ann/init_rules/gaussian_init.hpp"
#include "../../mlpack/src/mlpack/methods/ann/layer/layer.hpp"
#include "../../mlpack/src/mlpack/methods/ann/loss_functions/mean_squared_error.hpp"

#include "../../mlpack/src/mlpack/methods/reinforcement_learning/environment/acrobot.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/environment/env_type.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/environment/double_pole_cart.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/q_learning.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp"
#include "../../mlpack/src/mlpack/methods/reinforcement_learning/training_config.hpp"

#include <ensmallen.hpp>
#include <iostream>

#include "environment.hpp"
using namespace gym;

using namespace mlpack;
using namespace mlpack::ann;
using namespace ens;
using namespace mlpack::rl;

int main(int argc, char *argv[])
{
    mlpack::math::RandomSeed(std::time(NULL));
    DiscreteActionEnv::State::dimension = 6;
    DiscreteActionEnv::Action::size = 3;

    // Set up the network.
    SimpleDQN<> model(DiscreteActionEnv::State::dimension, 32, 32, DiscreteActionEnv::Action::size);

    // Set up the policy and replay method.
    GreedyPolicy<DiscreteActionEnv> policy(1.0, 1000, 0.1, 0.99);
    RandomReplay<DiscreteActionEnv> replayMethod(64, 5000, 3);

    TrainingConfig config;
    config.ExplorationSteps() = 400;

    const std::string environment = "Acrobot-v1";
    const std::string host = "127.0.0.1";
    const std::string port = "4040";

    Environment env(host, port, environment);
    // Set up DQN agent.
    QLearning<DiscreteActionEnv, decltype(model), AdamUpdate, decltype(policy), decltype(replayMethod)>
        agent(config, model, policy, replayMethod);

    std::vector<double> returnList;
    size_t episodes = 0;
    bool converged = true;
    size_t consecutiveEpisodesTest = 80;
    while (true)
    {
        double episodeReturn = 0;
        env.reset();
        do
        {
            agent.State().Data() = env.observation;
            agent.SelectAction();
            arma::mat action = {double(agent.Action().action)};
            // arma::mat action = env.action_space.sample();

            env.step(action);
            DiscreteActionEnv::State nextState;
            nextState.Data() = env.observation;

            replayMethod.Store(agent.State(), agent.Action(), env.reward, nextState, env.done, 0.99);
            episodeReturn += env.reward;
            agent.TotalSteps()++;
            if (agent.Deterministic() || agent.TotalSteps() < config.ExplorationSteps())
                continue;
            agent.TrainAgent();
        } while (!env.done);
        returnList.push_back(episodeReturn);
        episodes += 1;

        if (returnList.size() <= consecutiveEpisodesTest)
        {
            std::cout << episodeReturn << std::endl;
            continue;
        }
        else
            returnList.erase(returnList.begin());
        double averageReturn = std::accumulate(returnList.begin(),
                                               returnList.end(), 0.0) /
                               returnList.size();

        std::cout << "Average return in last " << consecutiveEpisodesTest
                  << " consecutive episodes: " << averageReturn
                  << " Episode return: " << episodeReturn << std::endl;

        if (episodes > 1000)
        {
            std::cout << "Cart Pole with DQN failed." << std::endl;
            converged = false;
            break;
        }
        if (averageReturn > -120)
            break;
    }
    env.compression(9);
    env.monitor.start("./dummy/", true, true);

    env.reset();
    env.render();

    for (int i = 0; i < 50; i++)
    {
        double currentReward = 0;
        size_t currentStep = 0;
        env.reset();
        while (1)
        {
            agent.State().Data() = env.observation;
            agent.SelectAction();
            arma::mat action = {double(agent.Action().action)};
            // arma::mat action = env.action_space.sample();

            env.step(action);
            currentReward += env.reward;
            currentStep += 1;

            if (env.done)
            {
                std::cout << "current reward: " << currentReward << " iteration: " << i << std::endl;
                break;
            }

            // std::cout << "Current step: " << currentStep << " current reward: "
            //           << currentReward << std::endl;
        }
    }

    std::cout << "Video: https://kurg.org/media/gym/" << env.instance
              << " (it might take some minutes before the video is accessible)."
              << std::endl;

    return 0;
}
```

Could you try this out locally. It's able to get the same rewards on test(aka rendered env), as it got during training. And the model is able to converge for all runs I ran on my laptop.(using random seeds).

nishantkr18:
Shall we use one test for acrobot in the notebook? Because here, its able to converge and show a solved render in every run.

zoq:
> > Did some further testing looks like the result is better if you increase the model size and use `RandomReplay` but it's not consistently, at least I couldn't produce a model to worked in each run. Do you think we should test some other env and see if the agent is able to solve it?
> 
> Yeah, let me go ahead and do that. Btw, did you run the tests online on notebook or locally?

Online

> Shall we use one test for acrobot in the notebook? Because here, its able to converge and show a solved render in every run.

Not sure what you mean, can you elaborate?

nishantkr18:
> Not sure what you mean, can you elaborate?

We use 2 tests now right ? One after training for some steps, and one after final training. Instead could we train completely all at once and then test the agent only once at the end? Doing this generally renders a solved environment in the end.
I've added what I said in the last commit, could you hve a look?

zoq:
Thanks for the clarification, sounds reasonable, I tested the updated notebook and it worked most of the time, maybe we should increase the number of training steps to make sure the agent has enough time to learn the env, had some good results with: `30000`.

nishantkr18:
> Thanks for the clarification, sounds reasonable, I tested the updated notebook and it worked most of the time, maybe we should increase the number of training steps to make sure the agent has enough time to learn the env, had some good results with: `30000`.

I trained the agent for 30000 steps, and the result is https://gist.github.com/nishantkr18/81473b087ba21759e099a576402204b2 Here, I've just copy-pasted the last cell thrice. So there are 4 consequitive tests, along with their renders. Notice how the 1st render gives `-500` while the others give around `-100`.

zoq:
Also tested:

```c++
// Set up the network.
FFN<MeanSquaredError<>, RandomInitialization> module(MeanSquaredError<>(), RandomInitialization(-1, 1));
    module.Add<Linear<>>(DiscreteActionEnv::State::dimension, 64);
    module.Add<ReLULayer<>>();
    //module.Add<Linear<>>(64, 32);
    //module.Add<ReLULayer<>>();
    module.Add<Linear<>>(64, DiscreteActionEnv::Action::size);


SimpleDQN<FFN<MeanSquaredError<>, RandomInitialization>> model(module);
```

with 1000 training steps, which looks quite more stable.

zoq:
> This means that maybe testing the agent once isn't enough 😄

Hm, maybe it's a combination of network size and initial weights.

zoq:
Also, looks like the SimpleDQN doesn't use the template parameter to construct the network: https://github.com/mlpack/mlpack/blob/ed3a774d624ebcac2a3b9091a873fc1950cf4987/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp#L29-L30

nishantkr18:
> with 1000 training steps, which looks quite more stable.

Tried it, results are promising.

> Hm, maybe it's a combination of network size and initial weights.

But I'm not sure that while testing, why does a "trained agent" performs well in most tests, but dreadfully fails in one of them. 

> Also, looks like the SimpleDQN doesn't use the template parameter to construct the network: https://github.com/mlpack/mlpack/blob/ed3a774d624ebcac2a3b9091a873fc1950cf4987/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp#L29-L30

Hmm :thinking: are u sure? I think we use it [here](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp#L75) and [here](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp#L54). Maybe I missunderstood something?

zoq:
Right, the problem is I can't use the constructor that allows me to specify the layer dimension with another intialization strategy other than `GaussianInitialization`. Because https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp#L54 doesn't use `NetworkType()`.

zoq:
> But I'm not sure that while testing, why does a "trained agent" performs well in most tests, but dreadfully fails in one of them.

Since the env starts in a random state, it's quite possible that the agent is not able to solve it in each case, we could increase the number of training steps and see if that helps. Let me know what you think.

nishantkr18:
> Right, the problem is I can't use the constructor that allows me to specify the layer dimension with another intialization strategy other than `GaussianInitialization`. Because https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp#L54 doesn't use `NetworkType()`.

Oh, got it now. I've opened https://github.com/mlpack/mlpack/pull/2540, I think it should solve the issue. please have a look.

> Since the env starts in a random state, it's quite possible that the agent is not able to solve it in each case, we could increase the number of training steps and see if that helps. Let me know what you think.

I'm not so sure if that works. I had tried different training step values, till `30000`, all of them showed similar performance when it came to testing (i.e. sometimes randomly one test just fails miserabely, while testing the same agent again yeilds a better result).
In this case, the I think we can add two tests in the end, instead of one. And then add a small note explaining that "the agent may sometimes not be able to perform well, so we test twice just to be sure. You may also test the agent any no of times by rerunning the testing cell". What do u think?

zoq:
> In this case, the I think we can add two tests in the end, instead of one. And then add a small note explaining that "the agent may sometimes not be able to perform well, so we test twice just to be sure. You may also test the agent any no of times by rerunning the testing cell". What do u think?

Sounds like a good plan to me.

q_learning/acrobot_dqn.ipynb
```diff
@@ -0,0 +1,504 @@
...
+    "size_t consecutiveEpisodes = 50;"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Function to train the agent on mlpack's own implementation of the CartPole environment.\n",
```
zoq, line 207: I guess we should go with `Acrobot-v1` here and also sounds like the "own implementation" is an old statement from the previous version of the notebook.

q_learning/acrobot_dqn.ipynb
```diff
@@ -0,0 +1,504 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Set up the network.\n",
+    "FFN<MeanSquaredError<>, RandomInitialization> module(MeanSquaredError<>(), RandomInitialization(-1, 1));\n",
```
zoq, line 131: What do you think, should we wait for https://github.com/mlpack/mlpack/pull/2540 and adjust this here or should we go with the network instantiation used here.

nishantkr18: I'm ok with both :)

q_learning/acrobot_dqn.ipynb
nishantkr18: Oh, missed that. let me do the changes..

q_learning/acrobot_dqn.ipynb
nishantkr18: If this PR is ready, lets merge it?

favre49: Personally, I think it would make for a better example to show flexibility through template arguments after we merge #2540? In general showing a variety of different approaches we can use is good for examples. Maybe we could show that in a different notebook as well, just something to keep in mind I suppose

zoq: Good idea, so let's keep it.

q_learning/acrobot_dqn.ipynb
```diff
@@ -0,0 +1,504 @@
...
+    "    module.Add<Linear<>>(DiscreteActionEnv::State::dimension, 64);\n",
+    "    module.Add<ReLULayer<>>();\n",
+    "    module.Add<Linear<>>(64, DiscreteActionEnv::Action::size);\n",
```
zoq, line 134:
```suggestion
    "module.Add<Linear<>>(DiscreteActionEnv::State::dimension, 64);\n",
    "module.Add<ReLULayer<>>();\n",
    "module.Add<Linear<>>(64, DiscreteActionEnv::Action::size);\n",
```

Let's remove the extra space here.

q_learning/acrobot_dqn.ipynb
```diff
@@ -0,0 +1,504 @@
...
+    "video"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Due to the stochasticity of the environment, it's quite possible that sometimes the agent is not able to solve it in each test. So, we test the agent once more, just to be sure.\n",
+    "\n",
+    "You may test the agent any no of times by rerunning either of the testing cells. "
```
favre49, line 406:
```suggestion
    "You may test the agent any number of times by rerunning either of the testing cells. "
```

PR #101, opened 2020-07-21
PCA examples

zoq:
Two example notebooks for data transformation using PCA; the reason I went with two examples is that I like to show that mlpack is able to work with csv and image data. Currently the examples are implemented in Python, I will add notebooks for the other languages, once this passed a first review.

Live: https://lab.mlpack.org/v2/gh/zoq/examples/pca-example?urlpath=lab%2Ftree%2Fbreast-cancer-wisconsin_transformation_with_pca%2Fbreast-cancer-wisconsin-pca-py.ipynb

PR #103, opened 2020-07-22
Adding notebook for solving mountainCar using DQN

nishantkr18:
I'm trying to use a custom reward function for training, since the original one will require a lot of episodes before an agent is able to do some successful training.

The reward function that I use here, is able to train the agent in pytorch implementation within 100 episodes. But here, the agent is not able to converge for even `300` episodes.

I've been training it on the notebook. Let me try locally and see.


nishantkr18:
With local training, and using the reward function:
```c++
adjustedReward = reward;
if (nextState.Data()[0] < -0.8)
    adjustedReward += 0.5;
```
the agent is able to converge in around 600 episodes.

nishantkr18:
I also noticed that the computation time for each episode in notebook is around 1 sec, whereas locally around 5 episodes take place in one second. If we could decrease the time for computation for each episode, we would be able to train the agent faster and without any custom reward function, IMO.

q_learning/mountain_car_dqn.ipynb
```diff
@@ -0,0 +1,814 @@
+{
```
favre49, line 1: <p>I think you'll need to update the description here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/103/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

favre49:
You're right, I'll also try testing this to see if we can change any parameters to make it run faster or converge sooner

zoq:
> I also noticed that the computation time for each episode in notebook is around 1 sec, whereas locally around 5 episodes take place in one second. If we could decrease the time for computation for each episode, we would be able to train the agent faster and without any custom reward function, IMO.

Can you test https://gist.github.com/zoq/7c1f8d29e7ec8229b67acdaa608e265a locally as well? Wondering if the Erlang package handler is faster.

But I guess cling is the limiting factor here, not much we can do about it. Also not sure I see the connection between runtime and reward. "If we could decrease the time for computation for each episode, we would be able to train the agent faster and without any custom reward function, IMO." Correct me if I'm wrong, but currently we don't use a custom reward function?

zoq:
Ahh, sorry looked at the wrong code, still I don't see the connection between runtime and custom reward function.

nishantkr18:
> still I don't see the connection between runtime and custom reward function.

Oh, I meant that if computation time is reduced, then we can run many episodes fast, so we can get rid of custom reward function, since we use custom rewards just to ensure earlier(and therefore faster) training of the agent. 

I say this because on locally when i train, the agent converges withen 1.5 mins of training. (around 600to800 episodes, without custom reward function)

q_learning/mountain_car_dqn.ipynb
nishantkr18: Yup, I'll do that once we're almost ready. Thanks for reminding :)

zoq:
I see, if the performance of the python server is a problem I can see if I can optimize some lines, but if it's cling well I think we just have to go with the slower training. So if you can rerun your timing test with the python server that would be great.

nishantkr18:
> I see, if the performance of the python server is a problem I can see if I can optimize some lines, but if it's cling well I think we just have to go with the slower training. So if you can rerun your timing test with the python server that would be great.

Yeah, I just ran the test with server.py file, the timing is the same. 1.5 mins to train the agent. But I'm not able to see the rendering of the trained agent in the end. I'm guessing that's some local issue maybe.

favre49:
I think even with the adjusted reward function, its very slow to converge. I've ran it a couple times with varying number of episodes and even above 1200 I'm yet to see the average reward drop below 130.

zoq:
> > I see, if the performance of the python server is a problem I can see if I can optimize some lines, but if it's cling well I think we just have to go with the slower training. So if you can rerun your timing test with the python server that would be great.
> 
> Yeah, I just ran the test with server.py file, the timing is the same. 1.5 mins to train the agent. But I'm not able to see the rendering of the trained agent in the end. I'm guessing that's some local issue maybe.

I meant to test it locally, I guess for your tests you used the local gym server as well? About the video, yes, you either print the url at the end, or add the changes I did here: https://github.com/mlpack/examples/pull/101/commits/d74adba1ef762d9eebf70c844094f8039391e32f to the binder postBuild config.

nishantkr18:
> I think even with the adjusted reward function, its very slow to converge. I've ran it a couple times with varying number of episodes and even above 1200 I'm yet to see the average reward drop below 130.

Do u mean 1200 steps? and average reward go above -130 ?

> I meant to test it locally, I guess for your tests you used the local gym server as well? About the video, yes, you either print the url at the end, or add the changes I did here: d74adba to the binder postBuild config.

By local gym server, do u mean starting the gym_tcp_api server using `iex -S mix`? Just to be clear, I ran the test using server.py and running the executable file, without anything else running.

nishantkr18:
> I think even with the adjusted reward function, its very slow to converge.

Yeah, that's the case here, but in pytorch implementation, with the custom reward function, the agent starts converging at 100 episodes, but here that happens at 500 episodes or sometimes more.

zoq:
> By local gym server, do u mean starting the gym_tcp_api server using `iex -S mix`? Just to be clear, I ran the test using server.py and running the executable file, without anything else running.

Maybe I misunderstood "I also noticed that the computation time for each episode in notebook is around 1 sec, whereas locally around 5 episodes take place in one second.", this was with `iex -S mix` running locally right? So if you say the timings are about the same I guess this is referring to `server.py` and `iex -S mix` running locally?

zoq:
> Yeah, that's the case here, but in pytorch implementation, with the custom reward function, the agent starts converging at 100 episodes, but here that happens at 500 episodes or sometimes more.

With the same network and optimizer settings?

nishantkr18:
> > By local gym server, do u mean starting the gym_tcp_api server using `iex -S mix`? Just to be clear, I ran the test using server.py and running the executable file, without anything else running.
> 
> Maybe I misunderstood "I also noticed that the computation time for each episode in notebook is around 1 sec, whereas locally around 5 episodes take place in one second.", this was with `iex -S mix` running locally right? So if you say the timings are about the same I guess this is referring to `server.py` and `iex -S mix` running locally?

Yes, you are absolutely correct. I should have said more clearly. I see how confusing it it now :sweat_smile: . Yes, the timings are about the same for `server.py` and `iex -S mix` running locally.


> > Yeah, that's the case here, but in pytorch implementation, with the custom reward function, the agent starts converging at 100 episodes, but here that happens at 500 episodes or sometimes more.
> 
> With the same network and optimizer settings?

Yes.

zoq:
> Yes, you are absolutely correct. I should have said more clearly. I see how confusing it it now 😅 . Yes, the timings are about the same for `server.py` and `iex -S mix` running locally.

Thanks for the clarification, now everything is clear, so this should be a cling issue.



> > > Yeah, that's the case here, but in pytorch implementation, with the custom reward function, the agent starts converging at 100 episodes, but here that happens at 500 episodes or sometimes more.
> > 
> > 
> > With the same network and optimizer settings?
> 
> Yes.

If we assume the implementation is the same, our initial weights are bad; we can test if `RandomInitialization` instead of `GaussianInitialization` does make a difference - https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/q_networks/simple_dqn.hpp#L30

favre49:
> Do u mean 1200 steps?

As in, 1200*200 steps

> and average reward go above -130 ?

Right, my bad. I'm testing this online, and I'm not sure why I'm yet to see it solve the problem

zoq:
> Right, my bad. I'm testing this online, and I'm not sure why I'm yet to see it solve the problem

Maybe that was my fault, I had to restart the service.

favre49:
Actually, I think it was a misunderstanding on my part. Like Nishant said, it does start converging after 500ish episodes, I'm not sure DQN is actually capable of completely solving mountain car (I think I saw Nishant comment about it before?)

nishantkr18:
Well, after training with the custom reward, here's the result :star_struck:  https://gym.kurg.org/1852fc9c85d44/output.webm

You were right @zoq, Initialization was a problem. I was using the same 0.001 as std for gaussian init. Changing it to 1 caused the agent to start training within 75 episodes with custom reward setting, just like the pytorch implementation. This surely made me realize the importance of initializing network paramteres.

Also, thanks for the update. Now each episode takes much less time!

nishantkr18:
> Actually, I think it was a misunderstanding on my part. Like Nishant said, it does start converging after 500ish episodes, I'm not sure DQN is actually capable of completely solving mountain car (I think I saw Nishant comment about it before?)

Well, in the official leaderboard, there are people who have used SARSA (some have used dqn) to solve mountain car, without any custom reward function, but they use ~650 episodes to get good scores.

nishantkr18:
With this ready, I was thinking of trying out pendulum(a continuous action space env) with dqn, by "discretizing" the action space. However, I dont think `env.step(action)` works for continuous action space envs. Is that correct?

zoq:
> I dont think `env.step(action)` works for continuous action space envs. Is that correct?

Do you mean the gym wrapper?

q_learning/mountain_car_dqn.ipynb
```diff
@@ -0,0 +1,588 @@
...
+    "This minor tweak can greatly increase sample efficiency."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Function to train the agent on mlpack's own implementation of the CartPole environment.\n",
```
zoq, line 216: Should this be `MountainCar`? Also maybe we should remove the "own implementation" part, since we use the gym implementation, I think this is a left over from the previous notebook.

nishantkr18: Yeah sure.

nishantkr18:
> > I dont think `env.step(action)` works for continuous action space envs. Is that correct?
> 
> Do you mean the gym wrapper?

Yeah.

zoq:
Since a continuous action space is represented as `gym.spaces.Box` it should be covered with https://github.com/zoq/gym_tcp_api/blob/master/priv/worker.py#L121. But if you run into any problem let me know and I will look into it.

nishantkr18:
> Since a continuous action space is represented as `gym.spaces.Box` it should be covered with https://github.com/zoq/gym_tcp_api/blob/master/priv/worker.py#L121. But if you run into any problem let me know and I will look into it.

It seems that continuous action space is not defined here: https://github.com/zoq/gym_tcp_api/blob/master/cpp/messages.hpp#L86

zoq:
> > Since a continuous action space is represented as `gym.spaces.Box` it should be covered with https://github.com/zoq/gym_tcp_api/blob/master/priv/worker.py#L121. But if you run into any problem let me know and I will look into it.
> 
> It seems that continuous action space is not defined here: https://github.com/zoq/gym_tcp_api/blob/master/cpp/messages.hpp#L86

Ohh you are right, I was looking into the wrong file. Do you want to take a look at adding the support or should I, don't feel obligated?

nishantkr18:
> > > Since a continuous action space is represented as `gym.spaces.Box` it should be covered with https://github.com/zoq/gym_tcp_api/blob/master/priv/worker.py#L121. But if you run into any problem let me know and I will look into it.
> > 
> > 
> > It seems that continuous action space is not defined here: https://github.com/zoq/gym_tcp_api/blob/master/cpp/messages.hpp#L86
> 
> Ohh you are right, I was looking into the wrong file. Do you want to take a look at adding the support or should I, don't feel obligated?

Ok,  let me try. I'll ask on irc if I face any problems :)

nishantkr18:
With this: https://gist.github.com/nishantkr18/61289882ec5e704ab7bd6c1bc3ee9e34 , I'm able to balance a pendulum using dqn, just using 3 actions{0, 1 ,-1} out of the range!

zoq:
Great, let's open another PR.

PR #104, opened 2020-07-22
Use 0 to numClass Labels instead of 1 to 10 where NLL is used.

kartikdutt18:
Hey everyone,
This PR makes the change from 0 to numClasses for training labels. The PR will be merged after mlpack/mlpack#2534. This PR also adds BatchNorm for ConvLayer to give better output and faster convergence.
Kindly let me know your thoughts.

Regards,
Kartik.

### DIFF
--- a/mnist_cnn/mnist_cnn.cpp
+++ b/mnist_cnn/mnist_cnn.cpp
@@ -79,13 +79,9 @@ int main()
   const mat trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1);
   const mat validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1);
 
-  // According to NegativeLogLikelihood output layer of NN, labels should
-  // specify class of a data point and be in the interval from 1 to
-  // number of classes (in this case from 1 to 10).
-
   // Create labels for training and validatiion datasets.
-  const mat trainY = train.row(0) + 1;
-  const mat validY = valid.row(0) + 1;
+  const mat trainY = train.row(0);
+  const mat validY = valid.row(0);
 
   // Specify the NN model. NegativeLogLikelihood is the output layer that
   // is used for classification problem. RandomInitialization means that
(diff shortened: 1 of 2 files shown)


kartikdutt18:
Yes, with minibatchnorm,  the model performs better and converges faster(as it should). Here are the results 
With BatchNorm

Screenshot 2020-07-05 at 4 03 00 PM

Without BatchNorm

Screenshot 2020-07-05 at 4 06 16 PM

rcurtin:
Awesome!  Note that it might be a while until mlpack 4.0.0 is released, so this might sit for a bit.  You might want to cherry-pick the non-NLL-related changes into a different PR that we can merge now---up to you. :+1:

rcurtin:
(I also will avoid approving it until 4.0.0 is released, so that we don't accidentally merge it and break things. :))

mnist_cnn/mnist_cnn.cpp
```diff
@@ -79,13 +79,9 @@ int main()
   const mat trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1);
   const mat validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1);
 
-  // According to NegativeLogLikelihood output layer of NN, labels should
-  // specify class of a data point and be in the interval from 1 to
-  // number of classes (in this case from 1 to 10).
```
rcurtin, line 84: It could be useful to keep this comment but adapt it such that it reads from 0 to `numClasses - 1` or something.  I think it's useful to keep because it indicates that the responses *shouldn't* be one-hot encoded (which is the default for some other toolkits).

kartikdutt18: Makes sense, I'll re-add it.

rcurtin:
Wow, I just now noticed this... it's the same changes I did in #142.  I could have saved myself a minute if I looked first!  Sorry @kartikdutt18 :(  Anyway, I'll go ahead and close this because the changes have now been done. :+1:

PR #106, opened 2020-07-27
Add BatchNorm Layer to mnist_cnn.

### LINKED ISSUE #105: MNIST CNN EXAMPLE NOT RUNNING PROPERLY
https://github.com/mlpack/examples/blob/master/mnist_cnn/mnist_cnn.cpp

The above code gives error while executiing it! I figured out that the problem was due to `mlpack::ann::NegativeLogLikelohood<>` so I replaced it with `mlpack::ann::MeanSquaredError`. The code runs fine now but the model isn't learning anything! I tried changing the model, learning rate, normalizing the data. Nothing works... Sometimes the error reduces to a very small number like 0.00415 but the accuracy (both test and valid) is always low (around 9-11%). I am trying for about a week to get the model working properly! Please help!

kartikdutt18:
Hey everyone, This PR adds BatchNorm Layer to mnist_cnn example.
Closes #105.
The results of training are given below :

With BatchNorm

Screenshot 2020-07-05 at 4 03 00 PM

Accuracy : 
```
Accuracy: train = 89.2%,         valid = 89.0482%
```
Without BatchNorm

Screenshot 2020-07-05 at 4 06 16 PM

Kindly let me know what you think.
Regards.

### DIFF
--- a/mnist_cnn/mnist_cnn.cpp
+++ b/mnist_cnn/mnist_cnn.cpp
@@ -45,8 +45,8 @@ int main()
   // and training parts with following ratio.
   constexpr double RATIO = 0.1;
 
-  // Allow infinite number of iterations until we stopped by EarlyStopAtMinLoss
-  constexpr int MAX_ITERATIONS = 0;
+  // Allow at max 10 iterations.
+  constexpr int MAX_ITERATIONS = 10;
 
   // Step size of the optimizer.
   constexpr double STEP_SIZE = 1.2e-3;
@@ -76,8 +76,10 @@ int main()
 
   // The train and valid datasets contain both - the features as well as the
   // class labels. Split these into separate mats.
-  const mat trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1);
-  const mat validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1);
+  const mat trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1) / 255;
+  const mat validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1) / 255;
+
+  
 
   // According to NegativeLogLikelihood output layer of NN, labels should
   // specify class of a data point and be in the interval from 1 to
@@ -123,6 +125,9 @@ int main()
                            28  // Input height.
   );
 
+  // Add BatchNorm layer.
+  model.Add<BatchNorm<>>(6, 1e-8, false);
+
   // Add first ReLU.
   model.Add<LeakyReLU<>>();
 
@@ -146,6 +151,9 @@ int main()
                            12  // Input height.
   );
 
+  // Add BatchNorm layer.
+  model.Add<BatchNorm<>>(16, 1e-8, false);
+
   // Add the second ReLU.
   model.Add<LeakyReLU<>>();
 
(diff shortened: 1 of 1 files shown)


mnist_cnn/mnist_cnn.cpp
```diff
@@ -45,8 +45,8 @@ int main()
   // and training parts with following ratio.
   constexpr double RATIO = 0.1;
 
-  // Allow infinite number of iterations until we stopped by EarlyStopAtMinLoss
-  constexpr int MAX_ITERATIONS = 0;
+  // Allow at max 10 iterations.
+  constexpr int MAX_ITERATIONS = 10;
```
zoq, line 49: With the BatchNorm layer, do we reach the max number of iterations or does, `EarlyStopAtMinLoss` reuturn before that?

kartikdutt18: With BatchNorm Layer the model converges in 10-11 epochs. The `EarlyStopAtMinLoss` keeps it training, even though there isn't any improvement in training / validation loss. So I set max_iterations to 10. I can reduce the patience of `EarlyStopAtMinLoss` if that's a better solution.

zoq: Sounds like a good idea, we can leave maxIterations set to 10.

kartikdutt18:
Keep open.

kartikdutt18:
Hey @zoq, I think this is good to go.

mnist_cnn/mnist_cnn.cpp
rcurtin: Actually in this case `MAX_ITERATIONS` should be set to the number of points in the training set * 10, since `MAX_ITERATIONS` refers to the number of points seen, not the number of epochs. :+1:

mnist_cnn/mnist_cnn.cpp
```diff
@@ -76,8 +76,10 @@ int main()
 
   // The train and valid datasets contain both - the features as well as the
   // class labels. Split these into separate mats.
-  const mat trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1);
-  const mat validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1);
+  const mat trainX = train.submat(1, 0, train.n_rows - 1, train.n_cols - 1) / 255;
+  const mat validX = valid.submat(1, 0, valid.n_rows - 1, valid.n_cols - 1) / 255;
+
+
```
rcurtin, line 82:
```suggestion
```
No need for extra blank lines. :)

rcurtin (commented):
Hey @kartikdutt18, just noticed this one was still open!  I think basically everything is good to go here, so if you want to merge master in we can see if everything still works fine. :+1:

mnist_cnn/mnist_cnn.cpp
kartikdutt18: Hey, Sure. I will pull in the latest branch, commit the suggestions, test and revert.

shrit:
I will try to port this example manually to this repo.

geekypathak21:
Hey @shrit I am thinking of picking this up let me know if there is any progress will look for something else. I just wanted to start with something small :sweat_smile:

shrit:
This is already done, let us close it. Thanks @geekypathak21

PR #110, opened 2020-08-13
Adding SAC example for pendulum env

nishantkr18:
Maybe, the binder service needs to be restarted. I cannot include the sac header.

Also, I think we can rename the `q_learning` folder to a more suitable name, maybe `reinforcement_learning`, as Soft-AC does not fall under the category of q_learning, I suppose.


zoq:
> Maybe, the binder service needs to be restarted. I cannot include the sac header.

I have to rebuild the mlpack package against the latest master branch, will do that today.

> Also, I think we can rename the `q_learning` folder to a more suitable name, maybe `reinforcement_learning`, as Soft-AC does not fall under the category of q_learning, I suppose.

Yes, good idea, maybe we can also say `reinforcement_learning_gym` to make clear we apply RL to gym env's.

zoq:
Currently rebuilding the package.

zoq:
Okay, so due to some recent updates (https://github.com/mlpack/mlpack/pull/2543) regarding the go package, we have to disable the go mlpack package - https://github.com/zoq/examples/commit/08bf8169d950217192308f5910d0467d124729bd should do the trick.

q_learning/pendulum_sac.ipynb
```diff
@@ -0,0 +1,446 @@
...
+     "ename": "Interpreter Error",
+     "evalue": "",
+     "output_type": "error",
+     "traceback": [
+      "Interpreter Error: "
+     ]
+    }
+   ],
+   "source": [
+    "#include <mlpack/methods/ann/ffn.hpp>\n",
```
zoq, line 58: `#include <mlpack/methods/ann/loss_functions/empty_loss.hpp>` is missing.

q_learning/pendulum_sac.ipynb
```diff
@@ -0,0 +1,446 @@
...
+    "#include <mlpack/methods/reinforcement_learning/sac.hpp>\n",
```
zoq, line 59: `#include <mlpack/methods/ann/init_rules/gaussian_init.hpp>` is missing.

q_learning/pendulum_sac.ipynb
```diff
@@ -0,0 +1,446 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "// Set up Soft actor-critic agent.\n",
+    "SAC<Pendulum, decltype(qNetwork), decltype(policyNetwork), AdamUpdate>\n",
```
zoq, line 191: I think we want to use the gym wrapper class instead of `Pendulum`?

nishantkr18:
> Okay, so due to some recent updates (mlpack/mlpack#2543) regarding the go package, we have to disable the go mlpack package - zoq@08bf816 should do the trick.

The training process is super slow using this, for me. Is that the case for you as well? I'll update the notebook here, so please use that to test.

zoq:
> The training process is super slow using this, for me. Is that the case for you as well? I'll update the notebook here, so please use that to test.

Yes I can see the same result, I timed the train block using:

```c++
arma::wall_clock timer;
timer.tic();

train(5000);
    
double n = timer.toc();
std::cout << "number of seconds: " << n << std::endl;
```

```
number of seconds: 345.687
```

I also tested with adding:

```c++
omp_set_dynamic(0);
omp_set_num_threads(2);
```

after the using namespace section and got slightly better results: `number of seconds: 292.053`. I'm not sure the cling kernel can handle openmp instructions. We can try to reduce the network size to speed things up.

nishantkr18:
> after the using namespace section and got slightly better results: `number of seconds: 292.053`. I'm not sure the cling kernel can handle openmp instructions. We can try to reduce the network size to speed things up.

I've reduced the network size and decreased the no. of updates in an interval. Though the agent is not able to perform as well as the "discretized" dqn, but still I think we can manage. Or, maybe train for longer and see the results?

zoq:
> I've reduced the network size and decreased the no. of updates in an interval. Though the agent is not able to perform as well as the "discretized" dqn, but still I think we can manage. Or, maybe train for longer and see the results?

Yeah, let's train with more iterations, I don't mind to wait a little but longer, if the result is more stable.

zoq:
Here we go, thanks again.

PR #115, opened 2020-09-20
Update xwidgets and cling

zoq:
Due to some changes in the Jupyter notebook package the current xwidgets package doesn't work anymore. There is a new xwidgets version 0.20.2, unfortunately the conda xwidgets 0.20.2 package has a conflict with the latest cling package, so we have to build the package manually until the package is fixed.

- Update to the latest cling version.
- Install the latest xwidgets version.

### DIFF
--- a/binder/kernel.json
+++ b/binder/kernel.json
@@ -1,0 +1,10 @@
+{
+  "display_name": "C++14",
+  "argv": [
+      "/srv/conda/envs/notebook/bin/xcpp",
+      "-f",
+      "{connection_file}",
+      "-std=c++14", "-fopenmp", "-pthread"
+  ],
+  "language": "C++14"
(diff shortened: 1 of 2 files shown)


binder/kernel.json
```diff
@@ -0,0 +1,10 @@
+{
+  "display_name": "C++14",
```
zoq, line 2: In the past, we used a custom xeus-cling package, that already came with a custom `kernel.json` file which builds against `openmp`, since we now use the upstream version, we have to re-apply the change after we installed the package.

rcurtin (approved):
Looks good to me. Thanks for the fix! :+1:

rcurtin:
Thanks @zoq! :+1:

PR #124, opened 2020-10-13
Ensmallen version check

### LINKED ISSUE #117: Error building the mnist_cnn and mnist_simple example
When i try to build the mnist_cnn.cpp and the mnist_simple.cpp i get the following error which is same for both files.
the version of mlpack i have is: 3.2.2-3
```
mnist_cnn.cpp: In function ‘int main()’:
mnist_cnn.cpp:189:20: error: no matching function for call to ‘ens::EarlyStopAtMinLoss::EarlyStopAtMinLoss(main()::<lambda(const mat&)>)’
  189 |                   }));
      |                    ^
In file included from /usr/include/ensmallen.hpp:67,
                 from /usr/include/mlpack/methods/ann/ffn.hpp:33,
                 from mnist_cnn.cpp:20:
/usr/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp:31:3: note: candidate: ‘ens::EarlyStopAtMinLoss::EarlyStopAtMinLoss(size_t)’
   31 |   EarlyStopAtMinLoss(const size_t patienceIn = 10) :
      |   ^~~~~~~~~~~~~~~~~~
/usr/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp:31:35: note:   no known conversion for argument 1 from ‘main()::<lambda(const mat&)>’ to ‘size_t’ {aka ‘long unsigned int’}
   31 |   EarlyStopAtMinLoss(const size_t patienceIn = 10) :
      |                      ~~~~~~~~~~~~~^~~~~~~~~~~~~~~
/usr/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp:21:7: note: candidate: ‘constexpr ens::EarlyStopAtMinLoss::EarlyStopAtMinLoss(const ens::EarlyStopAtMinLoss&)’
   21 | class EarlyStopAtMinLoss
      |       ^~~~~~~~~~~~~~~~~~
/usr/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp:21:7: note:   no known conversion for argument 1 from ‘main()::<lambda(const mat&)>’ to ‘const ens::EarlyStopAtMinLoss&’
/usr/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp:21:7: note: candidate: ‘constexpr ens::EarlyStopAtMinLoss::EarlyStopAtMinLoss(ens::EarlyStopAtMinLoss&&)’
/usr/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp:21:7: note:   no known conversion for argument 1 from ‘main()::<lambda(const mat&)>’ to ‘ens::EarlyStopAtMinLoss&&’
make: *** [<builtin>: mnist_cnn.o] Error 1
```

Aakash-kaushik:
Added ensmallen version check and corrected some data file names. 
Closes #117

### DIFF
--- a/mnist_batch_norm/mnist_batch_norm.cpp
+++ b/mnist_batch_norm/mnist_batch_norm.cpp
@@ -24,7 +24,7 @@
 #include <ensmallen.hpp>
 #include <ensmallen_bits/callbacks/callbacks.hpp>
 
-#if ((ENS_VERSION_MAJOR < 2) || (ENS_VERSION_MINOR < 13))
+#if ((ENS_VERSION_MAJOR < 2) || ((ENS_VERSION_MAJOR == 2) && (ENS_VERSION_MINOR < 13)))
   #error "need ensmallen version 2.13.0 or later"
 #endif
 
--- a/mnist_batch_norm/mnist_batch_norm.cpp
+++ b/mnist_batch_norm/mnist_batch_norm.cpp
@@ -24,6 +24,10 @@
 #include <ensmallen.hpp>
 #include <ensmallen_bits/callbacks/callbacks.hpp>
 
+#if ((ENS_VERSION_MAJOR < 2) || (ENS_VERSION_MINOR < 13))
+  #error "need ensmallen version 2.13.0 or later"
+#endif
+
 using namespace mlpack;
 using namespace mlpack::ann;
 
--- a/mnist_cnn/mnist_cnn.cpp
+++ b/mnist_cnn/mnist_cnn.cpp
@@ -21,6 +21,10 @@
 
 #include <ensmallen.hpp>
 
+#if ((ENS_VERSION_MAJOR < 2) || (ENS_VERSION_MINOR < 13))
+  #error "need ensmallen version 2.13.0 or later"
+#endif
+
 using namespace mlpack;
 using namespace mlpack::ann;
 
@@ -62,7 +66,7 @@ int main()
 
   // The original file can be downloaded from
   // https://www.kaggle.com/c/digit-recognizer/data
-  data::Load("../data/train.csv", tempDataset, true);
+  data::Load("../data/mnist_train.csv", tempDataset, true);
 
   // The original Kaggle dataset CSV file has headings for each column,
   // so it's necessary to get rid of the first row. In Armadillo representation,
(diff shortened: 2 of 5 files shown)


mnist_batch_norm/mnist_batch_norm.cpp
```diff
@@ -24,6 +24,10 @@
 #include <ensmallen.hpp>
 #include <ensmallen_bits/callbacks/callbacks.hpp>
 
+#if ((ENS_VERSION_MAJOR < 2) || (ENS_VERSION_MINOR < 13))
```
zoq, line 27: We should use `&&` here right?

Aakash-kaushik: So i installed a version of ensmallen less than 2.14 so i can actually see the #if working but with `&&` it wasn't producing any errors.

zoq: That is strange, if I didn't miss anything, the current construct will also raise an error if the version is 3.10.

Aakash-kaushik: Yes i think that will be the case, but i turned up with this solution because `&&` wasn't working. Also just a small implementation note on what i did.
`
#if (major version check && minor version check)
`
This is what i did and do you think 
`
#if (minor version check && major version check)
`
Will work as expected ?

Aakash-kaushik: I think a check similar to [this](https://github.com/mlpack/ensmallen/blob/a37f49cb9b6e003f92abb100b9758779545d02b3/include/ensmallen.hpp#L32) might work, i will check and push it in the morning.

zoq: Yes, that should solve the issue.

Aakash-kaushik: @zoq fixed it and verified it by manually changing the version of ensmallen to `3.10`. Doesn't produces a error for that.

zoq (approved):
Thanks for putting this together, no further comments from my side.

Aakash-kaushik:
> Thanks for putting this together, no further comments from my side.

Thanks for reviewing it. :D

zoq:
Thanks for the contribution.

PR #125, opened 2020-10-16
Update the example with best coordinates callback

shrit:


rcurtin (approved):
Ahh, nice catch, we should definitely use this too.

rcurtin:
Also, should we also apply the same changes to the other examples?

shrit:
Yes, I will update other examples too.

rcurtin:
Thanks @shrit! :+1:

PR #126, opened 2020-10-16
Return the value from the Lambda.

shrit:
I think I have already forgotten to add the return to this lambda (sorry). This will resolve any further issue from this example


zoq (approved):
Nice, thanks for the patch.

rcurtin (approved):
Thanks! :+1:

PR #127, opened 2020-10-19
Fix paths in some examples.

rcurtin:
While playing with the examples, I noticed that the paths aren't always the same as what the data downloader script saves things as.  These simple changes fix that.


birm (approved):
good catch!

zoq (approved):
Nice.

PR #128, opened 2020-10-19
Attempt to run examples on Travis.

rcurtin:
I'm not yet sure how to run any of the notebook examples as part of the build, but let's see if this script works for the C++ examples.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,9 @@ install:
   # Download datasets.
   - pwd
   - pip install tqdm
-  - tools/download_data_set.py
+  - cd tools/
+  - download_data_set.py
+  - cd ../
   - ls data/
 
 script:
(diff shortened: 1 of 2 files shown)


zoq:
> I'm not yet sure how to run any of the notebook examples as part of the build, but let's see if this script works for the C++ examples.

I can handle the notebook part.

rcurtin:
Ok, I think this one is ready!  There are four examples that don't work right, but I'll open issues for each of these and we can fix them one by one.  I think that in many cases the fixes should be pretty straightforward. :+1:

birm (approved):
Yey Testing!

zoq (approved):
Thanks for putting this together.

rcurtin:
Awesome; hopefully this will be helpful for others adding new examples. :+1:

Issue #97, opened 2020-06-19
Binding Examples [Discussion]


birm:
I'm not confident if this is a good idea, but it would be interesting to have a simple example for each of the bindings. This could double as a more robust binding test, as well as a demonstration. It may, however, add unneeded complexity.
I'm not sure. Thoughts?

kartikdutt18:
I think it's a nice idea. Since the examples repo is not restricted to models, It would be a good idea to showcase various features of mlpack including the bindings.

rcurtin:
I agree, it would be really nice.  For now we just have the quickstart examples, but perhaps those could be additionally moved here and then expanded on.  Some of the (very old) mlpack tutorials also have binding usage examples, but they aren't always great examples.

abernauer:
Yeah when writing that example with R binding I tried to expand on the quick start example and fit it into a standard machine learning workflow. I think for newcomers to mlpack having an example with how to use a binding in data science workflow is helpful. I think it's a solid idea.

rcurtin:
@birm do you think there is anything else to do for this issue? It seems like everyone is agreed that binding examples would definitely be a good idea. :)

birm:
Should I edit this issue to include subtasks for each binding then?

rcurtin:
@birm up to you; if you want to keep it open until we have examples of each binding, that sounds good to me.  Or, it might be easier to open a new issue for it and tag it with `help-wanted` `good-first-issue`, so that newcomers can quickly get started and see what needs to be done, as opposed to scrolling through the discussion here.  Your call though. :)

PR #129, opened 2020-10-21
added install.R. runtime.txt, and R notebook

abernauer:
1. Added the R notebook
2. Updated the binder directory with a runtime.txt file containing the R version.
3. Updated the binder directory with a install.R file with packages to install.


binder/install.R
```diff
@@ -0,0 +1,6 @@
+install.packages("mlpack")
```
zoq, line 1: I might be wrong but I think mlpack isn't in CRAN - https://github.com/mlpack/mlpack/pull/2636. I guess the easiest workaround for now is to enable the R bindings in the conda package. On your local setup you manually build the R bindings?

abernauer (commented):
My bad the comments were pending.

binder/install.R
zoq: I guess since installing the necessary packages takes a while it makes sense to include the files in the conda package. Will update the package later and post an update here.

birm:
If unintentional, this works towards #97. :1st_place_medal: 
This is awesome, but I don't feel I can give an approval since I learned exactly enough R to marginally pass assignments where it was needed :upside_down_face:

abernauer:
@birm Yes sort of unintentional. Though I had fun writing the example and would love to adapt some of the other examples for the R bindings.

abernauer:
Pushed the commit addressing the comments by @zoq  :smile:

movie_lens_prediction_with_cf/movie-lens-cf-r.ipynb
```diff
@@ -0,0 +1,933 @@
+{
```
rcurtin, line 1: <p>I think it would be really great if you added a quick header comment describing what this example does.  You could use the exact same comment from the other notebooks.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

rcurtin, line 1: <p>The <code>ratings-only.csv.gz</code>  file is 62MB, so this is a fairly large example.  Maybe it would be good to mention that in case a user thinks that this notebook will only take a second to run? :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

rcurtin, line 1: <p>Do you think we should put this comment just before we load the data? Also, it's probably good to introduce the whole idea of the MovieLens 20M dataset first.  Marcus has a nice comment describing it in the C++ notebook. (Where 20M refers to the number of ratings.)</p><p> </p><p>Also some quick formatting changes might be helpful. For instance, "userId an integer id for the user" might be clearer as "<code>userId</code> , an integer id representing the user".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

rcurtin, line 1: <p><code>Let's</code>  :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

rcurtin, line 1: <p><code>Peek</code>  not <code>Peak</code>  :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

rcurtin, line 1: <p>It might be nice to add a comment at the start of this section saying that we'll make a word cloud from movie titles.  Also, I think it's okay to do it all in one cell---no need to be super verbose.  In my view the word cloud is a cool visualization but not the primary point of the example.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

rcurtin, line 1: <p>It could be nice to wrap some of the longer command lines just for readability.  I assume not everyone will have a very wide monitor and many people may be looking on their phones, which are pretty narrow. :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

rcurtin, line 1: <p>Ah, nice---if you want to use descriptive statistics, it's probably good to add in a comment that this is mlpack support.  Specifically we are using mlpack to compute some nice statistics about each dimension in our data, but here there is only one dimension (the rating itself), so we get some nice statistics on the ratings.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

rcurtin, line 1: <p>Above these cells it could be nice to add a comment block describing that we will train the model, that we want to do a split into a train/test set first, and we'll use mlpack tools for both of these tasks, etc. etc.  Basically the reader will be scrolling through, and it's really helpful for them if we can describe <em>why</em> we're doing something instead of making them try and figure out why we did it. :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

rcurtin, line 1: <p>Looks like maybe one more space is needed so the arguments to cf() align?  Or maybe that is just my browser.</p><p> </p><p>What does the <code># model</code>  comment mean?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

rcurtin, line 1: <p><code>let's</code>  :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

rcurtin, line 1: <p>I actually don't recognize a single one of these movies.  I see that in the Python notebook Marcus used user 2; maybe user 2 has less obscure taste? :)  (Marcus also printed all of the movies that the user had previously rated, so that the reader can see the correlation between the two.  That would probably also be helpful here.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/mlpack/examples/pull/129/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

rcurtin:
Hey @abernauer, thanks for taking the time to implement this!  It looks great.  I left a bunch of comments through ReviewNB; hopefully they're helpful.  I do have some slight (but not serious) concern about the CF model itself; I wonder if we need to tune it with different parameters so that the movies it shows are more common.  Probably the `rank` parameter is what would need to be tuned.  Maybe worth a shot to see what happens and if the results look qualitatively better or worse? :+1:

movie_lens_prediction_with_cf/movie-lens-cf-r.ipynb
zoq: Maybe we should switch to the small data set, as done for the the other notebook. We can use the same settings, but it's much faster. In `movie-lens-cf-py.ipynb` I used `https://lab.mlpack.org/data/MovieLens-small.zip` which is < 1MB.

zoq: Same I don't know any of the listed movies, maybe that means I should watch more movies, also the movies are quite random, don't really see a pattern :)

shrit:
@abernauer this is nice work, but I will have to close this PR, we did a major refactoring of this repository and I would doubt that it will be easy to solve all the merge conflicts. Therefore, if you are still interested feel free to open this PR again.

PR #133, opened 2020-11-02
CI add cereal dependency

zoq:
mlpack serialization builds on cereal, which is currently not part of the CI build pipeline.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,10 @@ before_install:
   - cd $TRAVIS_BUILD_DIR
   - wget https://github.com/USCiLab/cereal/archive/v1.3.0.tar.gz
   - tar -xvzpf v1.3.0.tar.gz
+  - cd cereal-1.3.0
+  - cmake -DCMAKE_INSTALL_PREFIX=/usr .
+  - make
+  - make install
 
   # Build mlpack.
   - cd $TRAVIS_BUILD_DIR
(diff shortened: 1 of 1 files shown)


rcurtin (approved):
Nice catch!  Looks good to me. :+1:

birm (changes requested):
Noted error "/usr/include/mlpack/prereqs.hpp:86:38: fatal error: cereal/archives/binary.hpp: No such file or directory"

.travis.yml
```diff
@@ -29,13 +29,18 @@ before_install:
   - sudo make install
   - cd ../
 
+  # Install cereal.
+  - cd $TRAVIS_BUILD_DIR
+  - wget https://github.com/USCiLab/cereal/archive/v1.3.0.tar.gz
+  - tar -xvzpf v1.3.0.tar.gz
```
rcurtin, line 35: Oops, I approved too early.  I had just assumed that it worked. :smile:  If you use CMake to install cereal into /usr (so `CMAKE_INSTALL_PREFIX=/usr` when you configure cereal, then it shouldn't be necessary to specify `CEREAL_INCLUDE_DIR` when configuring mlpack and the examples should build and run okay. :+1:

birm (approved):
It looks like $target_name isn't set for the cereal directory, probably because it's not actually a test. Probably a good idea to fix that, but it looks like the examples are testing ok.

rcurtin:
Everything looks good to me now; I think it's ready for merge. :+1:

PR #136, opened 2020-11-12
Remove unused go plot package to resolve binder build.

zoq:
The PR fix the current Jupyter binder build. I guess keeping the Jupyter binder build happy is a never ending story.

### DIFF
--- a/binder/postBuild
+++ b/binder/postBuild
@@ -11,12 +11,12 @@ jupyter labextension uninstall --no-build jupyter-offlinenotebook
 # Upgrade jupyterlab if necessary.
 conda install jupyterlab=2
 
-# Install necessaary kernel and debugger related dependency following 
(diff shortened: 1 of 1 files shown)


binder/postBuild
```diff
@@ -11,12 +11,12 @@ jupyter labextension uninstall --no-build jupyter-offlinenotebook
 # Upgrade jupyterlab if necessary.
 conda install jupyterlab=2
 
-# Install necessaary kernel and debugger related dependency following 
+# Install necessaary kernel and debugger related dependency following
```
rcurtin, line 14:
```suggestion
# Install necessary kernel and debugger related dependency following
```

rcurtin (approved):
I think we spend more time making all our dependencies work than we do making our own software work... :)

birm (approved):

PR #137, opened 2020-11-12
Changes in examples

heisenbuug:
Made change in `mnist_batch_norm.cpp` and `lstm_stock_prediction.cpp`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,12 @@ script:
         continue;
       fi
 
+      # TODO: the dataset cannot be loaded.	
+      if [ "$dir" == "mnist_vae_cnn" ];	
+      then	
+        continue;	
+      fi	
+
       cd $dir;
       target_name=`grep 'TARGET := ' Makefile | awk -F ' := ' '{ print $2 }'`;
       echo "Make target $target_name in directory $dir.";
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,18 +78,6 @@ script:
         continue;
       fi
 
-      # TODO: the dataset appears to be in the wrong place.
-      #if [ "$dir" == "lstm_stock_prediction" ];
-      #then
-      #  continue;
-      #fi
-
-      # TODO: the prediction data appears to be the wrong size.
-      #if [ "$dir" == "mnist_batch_norm" ];
-      #then
-      #  continue;
-      #fi
-
       # TODO: the dataset cannot be loaded.
       if [ "$dir" == "mnist_vae_cnn" ];
       then
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,12 +78,6 @@ script:
         continue;
       fi
 
-      # TODO: the dataset cannot be loaded.
-      if [ "$dir" == "mnist_vae_cnn" ];
-      then
-        continue;
-      fi
-
       cd $dir;
       target_name=`grep 'TARGET := ' Makefile | awk -F ' := ' '{ print $2 }'`;
       echo "Make target $target_name in directory $dir.";
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,16 +79,16 @@ script:
       fi
 
       # TODO: the dataset appears to be in the wrong place.
-      if [ "$dir" == "lstm_stock_prediction" ];
-      then
-        continue;
-      fi
+      #if [ "$dir" == "lstm_stock_prediction" ];
+      #then
+      #  continue;
+      #fi
 
       # TODO: the prediction data appears to be the wrong size.
-      if [ "$dir" == "mnist_batch_norm" ];
-      then
-        continue;
-      fi
+      #if [ "$dir" == "mnist_batch_norm" ];
+      #then
+      #  continue;
+      #fi
 
       # TODO: the dataset cannot be loaded.
       if [ "$dir" == "mnist_vae_cnn" ];
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,10 +79,10 @@ script:
       fi
 
       # TODO: the dataset appears to be in the wrong place.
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -79,16 +79,16 @@ script:
...
 
       # TODO: the dataset appears to be in the wrong place.
-      if [ "$dir" == "lstm_stock_prediction" ];
-      then
-        continue;
-      fi
+      #if [ "$dir" == "lstm_stock_prediction" ];
+      #then
+      #  continue;
+      #fi
```
zoq, line 85: Since it's fixed, let's remove the block.

.travis.yml
```diff
@@ -79,16 +79,16 @@ script:
...
 
       # TODO: the prediction data appears to be the wrong size.
-      if [ "$dir" == "mnist_batch_norm" ];
-      then
-        continue;
-      fi
+      #if [ "$dir" == "mnist_batch_norm" ];
```
zoq, line 88: Since it's fixed, let's remove the block.

zoq (commented):
You are awesome, thanks for fixing the examples.

mnist_batch_norm/mnist_batch_norm.cpp
```diff
@@ -186,7 +186,7 @@ int main()
   mlpack::data::Load("../data/mnist_test.csv", tempDataset, true);
 
   mat testX =
-      tempDataset.submat(0, 1, tempDataset.n_rows - 1, tempDataset.n_cols - 1);
+      tempDataset.submat(1, 0, tempDataset.n_rows - 1, tempDataset.n_cols - 1);
```
rcurtin, line 189: Nice catch! :+1:

rcurtin (approved):
Awesome, looks good!  Thank you for fixing this @heisenbuug!  :100:

I think we may have to open another issue about tuning the batch norm implementation though:

```
Accuracy: train = 63.2407%, valid = 63.6106%
```

63% is... pretty bad for MNIST... :)

heisenbuug:
> Awesome, looks good! Thank you for fixing this @heisenbuug! 💯
> 
> I think we may have to open another issue about tuning the batch norm implementation though:
> 
> ```
> Accuracy: train = 63.2407%, valid = 63.6106%
> ```
> 
> 63% is... pretty bad for MNIST... :)

I agree @rcurtin, also actually I wanted to ask why don't we provide a dataset pre-processing library which would contain function like `.info` and `.describe()` in pandas.

zoq (approved):
Looks good to me, thanks for the fixes.

zoq:
> I agree @rcurtin, also actually I wanted to ask why don't we provide a dataset pre-processing library which would contain function like `.info` and `.describe()` in pandas. Having such tools to go through the dataset will make the tuning of models easy as we will get a better understanding of datasets.

I like the idea, I usually use http://arma.sourceforge.net/docs.html#stats_fns and http://arma.sourceforge.net/docs.html#running_stat to get some further information.

rcurtin:
Thanks @heisenbuug! :+1:  I saw you worked on the MNIST VAE example too, but it looks like that one's still not working (we can handle that some other time).

heisenbuug:
> > I agree @rcurtin, also actually I wanted to ask why don't we provide a dataset pre-processing library which would contain function like `.info` and `.describe()` in pandas. Having such tools to go through the dataset will make the tuning of models easy as we will get a better understanding of datasets.
> 
> I like the idea, I usually use http://arma.sourceforge.net/docs.html#stats_fns and http://arma.sourceforge.net/docs.html#running_stat to get some further information.

Yea, I went through that as well, and I created something like this, actually, I was thinking we could use just using one function to get all the info, I was planning to use these functions in the background only...
Capture

```
class Data {
private:
  std::string path;
  std::vector<string> headers;
  arma::mat data;

public:
  Data(string file_path)
  {
    path = file_path;
    mlpack::data::Load(path, data, true);
    data = data.submat(0, 1, data.n_rows - 1, data.n_cols - 1);
    getHeaders();
  }

  void info()
  {
    cout << typeid(data).name() << endl;
    std::cout << "Index Range: " << data.n_cols << std::endl;
    std::cout << "Data Columns (total " << data.n_rows << "):" << std::endl;
    for (auto i : headers)
      cout << i << endl;
  }
  void getHeaders()
  {
    std::fstream fin;
    fin.open(path, std::ios::in);
    std::string line, word, temp;
    getline(fin, line);
    stringstream s(line);
    while (getline(s, word, ','))
    {
      headers.push_back(word);
    }
  }

  void infoHelper(int lower, int upper)
  {
    for (auto i : headers)
    {
      cout.width(10);
      cout << i << ", ";
    }

    std::cout << std::endl;
    cout.precision(2);
    cout.setf(ios::fixed);
    cout.width(10);
    data.submat(0, upper, data.n_rows - 1, lower).t().raw_print(cout);
  }

  void head()
  {
    infoHelper(4, 0);
  }

  void tail()
  {
    infoHelper(data.n_cols - 1, data.n_cols - 5);
  }
};
```

This is the code I implemented. Now to call it:
```
int main()
{ 
  Data dataset(../housing.csv");
  dataset.info();
  return 0;
}
```
Are we interested in such functionality? I think data pre-processing and analysis will come in handy for users to take a quick glance at datasets. Shall I open another issue for this @zoq @rcurtin ?

I am thinking of also implementing a `describe`(open to naming suggestions) method which will contain `count mean std min max` and other insightful information. Pandas' methods mostly inspire these and as @zoq suggested we already have many functions available.

rcurtin:
@heisenbuug yeah, definitely something like this can be useful, especially in a notebook environment.  Let's continue the discussion in mlpack/mlpack#2722. :+1:

PR #138, opened 2020-11-23
BUG corrections Recurrent Neural Network Examples

zsogitbe:
- lstm_stock_prediction.cpp
- lstm_electricity_consumption.cpp

BUG corrections:
- trainX, trainY and testX, testY sizes caused a memory leak
- SaveResults correction and adjustment to the comment above the function.


zoq:
Thanks for the fix, super important to have a correct example. I'm wondering why this one passed the ci pipeline.

zsogitbe:
Yes the examples are very important! It is probably not the kind of 'memory leak' that ci can catch. The arrays had 1 extra element which were not initialized and were used in the calculations.

rcurtin:
@zsogitbe thanks for the report and PR!  I spent a while looking at the code but I can't reason how this fixes an invalid memory access.  In fact the memory accesses to `testX` and `predictions` appear to be exactly the same, and the temporary storage of predictions in `temp` doesn't seem to change anything.  I ran the electricity prediction through valgrind (a memory checker) and there were no uses of uninitialized memory when computing the `SaveResults` or anything.

Can you provide a bit more detail about what is exactly wrong here?  At this point I can't see what's wrong.

zsogitbe:
> @zsogitbe thanks for the report and PR! I spent a while looking at the code but I can't reason how this fixes an invalid memory access. In fact the memory accesses to `testX` and `predictions` appear to be exactly the same, and the temporary storage of predictions in `temp` doesn't seem to change anything. I ran the electricity prediction through valgrind (a memory checker) and there were no uses of uninitialized memory when computing the `SaveResults` or anything.
> 
> Can you provide a bit more detail about what is exactly wrong here? At this point I can't see what's wrong.

Yes it is very difficult to see the problem in the original code. Basically the column size (data records) of the following is one more than needed: `arma::cube trainX, trainY, testX, testY;` and this caused a crash in Debug mode in the original code (this is how I have found the problem). In Release mode it was strangely working fine but the results were calculated with the last element included which was not initialized thus depending on the value of the memory location it gave a biased result.

rcurtin:
Can you point to the exact line of the original code that has the invalid access?

zsogitbe:
> Can you point to the exact line of the original code that has the invalid access?

let me explain it with an example:

```
trainData.n_cols = 2441 
rho = 10
```
In CreateTimeSeriesData function 
```
i_max = 2430 (the maximum of i as an index)
trainX.n_cols was set in the old (wrong code) to 2432 (2441 - 10 + 1)
trainX.n_cols is set in the new (corrected) code to 2431 (2441 - 10)
```
So in CreateTimeSeriesData the trainX is filled in till index 2430 which is a column size of 2431. If trainX is initialized to 2432 size (old wrong code) then it will have one more element which is not used and not initialized.

The answer to the question where this is used is simple, it is used everywhere where trainX (or trainY, etc.) appears. For example the code `MSE(predOutP, testY);` is `metric::SquaredEuclideanDistance::Evaluate(pred, Y) / (Y.n_elem); ` where all elements of Y are used, thus the last one also... etc.

rcurtin:
Thanks for the clarification. When I look at `CreateTimeSeriesData()`, it does appear to set one fewer column when compared with how `trainX` and other cubes are initialized.

But why not simply solve this issue by removing the `+ 1` from the `set_size()` calls on lines 214-217?  The changes proposed by the PR don't appear to actually solve the issue you are pointing out.

zsogitbe:
> Thanks for the clarification. When I look at `CreateTimeSeriesData()`, it does appear to set one fewer column when compared with how `trainX` and other cubes are initialized.
> 
> But why not simply solve this issue by removing the `+ 1` from the `set_size()` calls on lines 214-217? The changes proposed by the PR don't appear to actually solve the issue you are pointing out.

Yes that is exactly what I have done in the correction but I had to change the SaveResults function also because it was over-optimized (my original code was changed by someone with good intentions) and it did not correspond to the comments above the function and to the aim of the example (it replaced the two last columns of the input instead of adding the predicted two last columns at the end). This is the reason for the other changes. Both are corrections and are needed.

rcurtin:
Sorry, I must have been somehow looking at the wrong diff!  I never saw the `trainX` changes somehow.  Sorry about that---those look just fine.

I don't think the `SaveResults()` changes are valid. We cannot use `scale.InverseTransform()` unless it has the same dimension as the data used when `scale.Transform()` was called.  So we cannot call `InverseTransform()` with `temp`; it will be unscaled incorrectly.  However, looking at the existing code, it is actually incorrect also; we should use `insert_row()` to *add* a row for the predictions to `flatDataAndPreds` instead of overwriting the last column.

zsogitbe:
> Sorry, I must have been somehow looking at the wrong diff! I never saw the `trainX` changes somehow. Sorry about that---those look just fine.
> 
No problem. Happy to help!
> I don't think the `SaveResults()` changes are valid. We cannot use `scale.InverseTransform()` unless it has the same dimension as the data used when `scale.Transform()` was called. So we cannot call `InverseTransform()` with `temp`; it will be unscaled incorrectly. However, looking at the existing code, it is actually incorrect also; we should use `insert_row()` to _add_ a row for the predictions to `flatDataAndPreds` instead of overwriting the last column.

Do not forget that there is only one column of data!

rcurtin:
Thank you for pointing that out!  I'm sorry for my misunderstanding---I jump between quite a lot of requests... Indeed in this case calling `InverseTransform()` in this way will be fine.  I have some slight nitpicks about grammar that I will add but otherwise this looks fine.  *Thanks* for working out this bug!

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
-  // any dataset for prediction.
-  cout << "The predicted energy consumption for the next hour is : " << endl;
-  cout << " "
-       << flatDataAndPreds(flatDataAndPreds.n_rows - 1,
-                           flatDataAndPreds.n_cols - 1)
-       << endl;
+  arma::mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction result is the energy consumption for the next hour and
```
rcurtin, line 99:
```suggestion
  // The prediction result is the energy consumption for the next hour and 
```
I can't seem to do multiline suggestions on a phone... sorry that these are split up...

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
-  cout << "The predicted energy consumption for the next hour is : " << endl;
-  cout << " "
-       << flatDataAndPreds(flatDataAndPreds.n_rows - 1,
-                           flatDataAndPreds.n_cols - 1)
-       << endl;
+  arma::mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction result is the energy consumption for the next hour and 
+  //comming from the last slice of the prediction.
```
rcurtin, line 100:
```suggestion
  // comes from the last slice of the prediction.
```

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
+  arma::mat temp = predictions.slice(predictions.n_slices - 1);
+  scale.InverseTransform(temp, temp);
+
+  //We shift the predictions into place to be in syncrone with the input data
```
rcurtin, line 104:
```suggestion
  // We shift the predictions such that the true values are synchronized with
```

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
+  //and add one more record to the input. Please note that in this way the last
```
rcurtin, line 105:
```suggestion
  // the predictions, and we also add one more record to the input. Please note
```

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction result is the energy consumption for the next hour and 
+  //comming from the last slice of the prediction.
+  arma::mat temp = predictions.slice(predictions.n_slices - 1);
+  scale.InverseTransform(temp, temp);
+
+  //We shift the predictions into place to be in syncrone with the input data
+  //and add one more record to the input. Please note that in this way the last 
+  //input record is replaced with zeros and the first prediction record is also zero.
```
rcurtin, line 106:
```suggestion
  // that this means the last input record zero and the first prediction record
  // is also zero.
```

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
+  temp.insert_cols(0, 1, true);
+  flatDataAndPreds.insert_cols(flatDataAndPreds.n_cols, 1, true);
+
+  //we add the prediction as the last column
```
rcurtin, line 110:
```suggestion
  // We add the predictions as the last column.
```

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
+  flatDataAndPreds.insert_rows(flatDataAndPreds.n_rows, temp.rows(temp.n_rows - 1, temp.n_rows - 1));
+
+  //Save the data to file. The last columns are the predictions; the preceding
```
rcurtin, line 113:
```suggestion
  // Save the data to file. The last columns are the predictions; the preceding
```

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
+  //and add one more record to the input. Please note that in this way the last 
+  //input record is replaced with zeros and the first prediction record is also zero.
+  temp.insert_cols(0, 1, true);
+  flatDataAndPreds.insert_cols(flatDataAndPreds.n_cols, 1, true);
+
+  //we add the prediction as the last column
+  flatDataAndPreds.insert_rows(flatDataAndPreds.n_rows, temp.rows(temp.n_rows - 1, temp.n_rows - 1));
+
+  //Save the data to file. The last columns are the predictions; the preceding 
+  //column is the data used to generate those predictions.
```
rcurtin, line 114:
```suggestion
  // column is the data used to generate those predictions.
```

lstm_electricity_consumption/lstm_electricity_consumption.cpp
```diff
@@ -93,33 +93,30 @@ void SaveResults(const string& filename,
...
+  data::Save(filename, flatDataAndPreds);
+
+  //Print the output to screen.
```
rcurtin, line 117:
```suggestion
  // Print the output to screen.
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
-       << flatDataAndPreds(flatDataAndPreds.n_rows - 2,
-                           flatDataAndPreds.n_cols - 1)
-       << ", ";
-  cout << flatDataAndPreds(flatDataAndPreds.n_rows - 1,
-                           flatDataAndPreds.n_cols - 1)
-       << ")" << endl;
+  arma::mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction results are the (high, low) for the next day and comming from
```
rcurtin, line 99:
```suggestion
  // The prediction results are the (high, low) for the next day and come from
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
-                           flatDataAndPreds.n_cols - 1)
-       << ", ";
-  cout << flatDataAndPreds(flatDataAndPreds.n_rows - 1,
-                           flatDataAndPreds.n_cols - 1)
-       << ")" << endl;
+  arma::mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction results are the (high, low) for the next day and comming from 
+  //the last slice from the prediction.
```
rcurtin, line 100:
```suggestion
  // the last slice from the prediction.
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  arma::mat temp = predictions.slice(predictions.n_slices - 1);
+
+  //NOTE: We add 3 extra rows here in order to recreate the input data
```
rcurtin, line 103:
```suggestion
  // NOTE: We add 3 extra rows here in order to recreate the input data
```
Thanks for this comment.  If I had seen it in my initial review I would have had a lot fewer questions.  I must have clicked on the link wrong or something. :(

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
-       << ")" << endl;
+  arma::mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction results are the (high, low) for the next day and comming from 
+  //the last slice from the prediction.
+  arma::mat temp = predictions.slice(predictions.n_slices - 1);
+
+  //NOTE: We add 3 extra rows here in order to recreate the input data 
+  //structure used to transform the data. This is needed in order to be able
```
rcurtin, line 104:
```suggestion
  // structure used to transform the data. This is needed in order to be able 
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  arma::mat flatDataAndPreds = testX.slice(testX.n_slices - 1);
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction results are the (high, low) for the next day and comming from 
+  //the last slice from the prediction.
+  arma::mat temp = predictions.slice(predictions.n_slices - 1);
+
+  //NOTE: We add 3 extra rows here in order to recreate the input data 
+  //structure used to transform the data. This is needed in order to be able 
+  //to use the right scaling parameters for the specific column stock
```
rcurtin, line 105:
```suggestion
  // to use the right scaling parameters for the specific column stock 
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  scale.InverseTransform(flatDataAndPreds, flatDataAndPreds);
+
+  //The prediction results are the (high, low) for the next day and comming from 
+  //the last slice from the prediction.
+  arma::mat temp = predictions.slice(predictions.n_slices - 1);
+
+  //NOTE: We add 3 extra rows here in order to recreate the input data 
+  //structure used to transform the data. This is needed in order to be able 
+  //to use the right scaling parameters for the specific column stock 
+  //high, low).
```
rcurtin, line 106:
```suggestion
  // (high, low).
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  temp.insert_rows(0, 3, 0);
+  scale.InverseTransform(temp, temp);
+
+  //We shift the predictions into place to be in syncrone with the input data
```
rcurtin, line 110:
```suggestion
  // We shift the predictions such that the true values are synchronized with
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  //and add one more record to the input. Please note that in this way the last
```
rcurtin, line 111:
```suggestion
  // the predictions, and we also add one more record to the input. Please note
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  //NOTE: We add 3 extra rows here in order to recreate the input data 
+  //structure used to transform the data. This is needed in order to be able 
+  //to use the right scaling parameters for the specific column stock 
+  //high, low).
+  temp.insert_rows(0, 3, 0);
+  scale.InverseTransform(temp, temp);
+
+  //We shift the predictions into place to be in syncrone with the input data
+  //and add one more record to the input. Please note that in this way the last 
+  //input record is replaced with zeros and the first prediction record is also zero.
```
rcurtin, line 112:
```suggestion
  // that this means the last input record is zero and the first prediction record
  // is also zero.
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  temp.insert_cols(0, 1, true);
+  flatDataAndPreds.insert_cols(flatDataAndPreds.n_cols, 1, true);
+
+  //We add the prediction as the last two columns (stock high, low)
```
rcurtin, line 116:
```suggestion
  // We add the prediction as the last two columns (stock high, low).

```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  flatDataAndPreds.insert_rows(flatDataAndPreds.n_rows, temp.rows(temp.n_rows - 2, temp.n_rows - 1));
+
+  //Save the data to file. The last columns are the predictions; the preceding
```
rcurtin, line 119:
```suggestion
  // Save the data to file. The last columns are the predictions; the preceding 
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  //and add one more record to the input. Please note that in this way the last 
+  //input record is replaced with zeros and the first prediction record is also zero.
+  temp.insert_cols(0, 1, true);
+  flatDataAndPreds.insert_cols(flatDataAndPreds.n_cols, 1, true);
+
+  //We add the prediction as the last two columns (stock high, low)
+  flatDataAndPreds.insert_rows(flatDataAndPreds.n_rows, temp.rows(temp.n_rows - 2, temp.n_rows - 1));
+
+  //Save the data to file. The last columns are the predictions; the preceding 
+  //columns are the data used to generate those predictions.
```
rcurtin, line 120:
```suggestion
  // columns are the data used to generate those predictions.
```

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  data::Save(filename, flatDataAndPreds);
+
+  //Print the output to screen.
```
rcurtin, line 123:
```suggestion
  // Print the output to screen.
```

rcurtin (commented):
Wow, I don't know how to do multiline suggestions on my phone so this was super tedious and is probably pretty ugly to look at.  But... the suggestions should apply cleanly. :)  All the suggestions are just changes to comments to fix grammar and match mlpack style conventions. :+1:

lstm_stock_prediction/lstm_stock_prediction.cpp
```diff
@@ -93,38 +93,37 @@ void SaveResults(const string filename,
...
+  cout << "The predicted Google stock (high, low) for the last day is: " << endl;
+  cout << "  (" << flatDataAndPreds(flatDataAndPreds.n_rows - 2, flatDataAndPreds.n_cols - 1) << " ,";
+  cout << flatDataAndPreds(flatDataAndPreds.n_rows - 1, flatDataAndPreds.n_cols - 1) << ")" << endl;
```
rcurtin, line 126:
```suggestion
  cout << "The predicted Google stock (high, low) for the last day is: "
      << endl;
  cout << "  (" << flatDataAndPreds(flatDataAndPreds.n_rows - 2,
      flatDataAndPreds.n_cols - 1) << " ,";
  cout << flatDataAndPreds(flatDataAndPreds.n_rows - 1,
      flatDataAndPreds.n_cols - 1) << ")" << endl;
```
Just some line length fixes.

rcurtin (approved):
This sat for a while, but I definitely want to see it merged so I went ahead and applied the suggestions.  @zsogitbe thank you again for these fixes!  If you disagree with any of the suggestions I added, please feel free to comment or open another PR. :+1:

rcurtin:
Thanks @zsogitbe!

PR #123, opened 2020-10-11
Contact tracing clustering with DBSCAN.

zoq:
Contact tracing clustering with DBSCAN (C++).

PR #139, opened 2020-11-25
Recommit contact-tracing-dbscan-cpp.ipynb notebook

zoq:
For some reason, binder isn't picking up the latest notebook, so what this commit does is to recommit the notebook, from the diff I can see that it has the latest changes. Which is strange since it's the same notebook as:

https://raw.githubusercontent.com/mlpack/examples/da0227d7983de4469319dee569417da52569af43/contact-tracing-dbscan-cpp.ipynb

from the #123. Anyway, this should fix the problem.


rcurtin (approved):
Strange, hopefully it works this time. :)

birm (approved):

zoq:
Alright, now it works correctly, thanks everyone.

PR #140, opened 2020-11-29
Jupyter/Binder build cleanup

zoq:
The image Binder builds for the repository is currently 9.94GB, ideally we can bring this down, removing the conda cache at the end, brings us down to 8.84GB.

I'll see if I can find some other options to save us some space, but I think this a good start.


birm:
ReviewNB doesn't seem to like the notebook.

birm (approved):
Oh, reviewNB didn't like it because you just removed the file. My bad! Looks good.

zoq:
Thanks for the review.

PR #119, opened 2020-09-27
Try to add cpp files for a reinforcement learning examples

shrit:
This pull request tries to make reinforcement learning examples run without the need for cpp notebooks.
I am not sure if this should work, it is only an attempt to reach a working example on my machine.
@zoq I might need your help in understanding what is happening, or do I need to install the server directly on my machine.
This example compiles successfully.
 I have copied the entire `cpp` directory from the `gym_tcp_api` environment,  I was not able to install the `gym_tcp_api` since there is not installing target.

PR #143, opened 2021-01-05
Clean-up binder path and binaries

zoq:
Simple clean-up after we changed some paths in https://github.com/mlpack/examples/pull/119.


birm (approved):
Ah, good catch!

PR #147, opened 2021-02-27
Don't remove header row from MNIST data

rcurtin:
The MNIST data that we download and preprocess doesn't have a header row, so there's no need to remove one.  This also cleans up a little bit of code here and there.


zoq (approved):
Nice :+1:

birm (approved):

rcurtin:
Oops, there was a bug---I forgot to remove the labels from the test data, but I fixed that in 5667948.  All tests are passing now, so I'll go ahead and merge. :+1:

PR #151, opened 2021-04-25
Jupyter notebook env update

zoq:
Overhaul the current Jupyter notebook environment:

- Use new binder package management configs.
- Include Python/C++ files for additional plotting.
- Add xeus-specific header file to set up the notebook instead of adding it into `core.hpp.`

I still have to update the mlpack conda package, because the package build by conda-forge does not have image support.


movie_lens_prediction_with_cf/movie-lens-cf-cpp.ipynb
```diff
@@ -51,13 +51,13 @@
     },
     {
       "cell_type":"code",
-      "source":"#include <mlpack/core.hpp>\n#include <mlpack/core/data/split_data.hpp>\n\n#include <mlpack/methods/cf/decomposition_policies/regularized_svd_method.hpp>\n#include <mlpack/methods/cf/cf.hpp>\n\n#include <fstream>",
+      "source":"#include <mlpack/xeus-cling.hpp>\n\n#include <mlpack/core.hpp>\n#include <mlpack/core/data/split_data.hpp>\n\n#include <mlpack/methods/cf/decomposition_policies/regularized_svd_method.hpp>\n#include <mlpack/methods/cf/cf.hpp>\n\n#include <fstream>",
```
rcurtin, line 54: Should we add this file upstream? I'm concerned about people adding this include outside of a notebook environment, where it may not work.

If the xeus notebook environment defines some macros or something, we could add some conditional includes in `mlpack/core.hpp` or something.

rcurtin (commented):
Looks good, just a couple comments about the build

 - Does the style check job need to be updated in Jenkins? I think you have been working on these lately but I am not sure of the details.

 - I think for the Travis issue we shouldn't bother addressing it in this PR but I think the 'usual' MNIST download source can be unreliable.  It might be better to use mlpack.org/datasets; not sure though.

movie_lens_prediction_with_cf/movie-lens-cf-cpp.ipynb
zoq: The file is somewhat specific to the notebook env:

```c++
#pragma cling load("/srv/conda/envs/notebook/lib/libarmadillo.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libmlpack.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_serialization.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_unit_test_framework.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_program_options.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_system.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_system.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_thread.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_iostreams.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libboost_filesystem.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libomp.so")
#pragma cling load("/srv/conda/envs/notebook/lib/libpython3.so")
#pragma cling add_include_path("/srv/conda/envs/notebook/include/python3.7m/")
```

I guess we could do some cmake magic to get all the files, but right now it's super specific to the setup we use. But let's open an issue, and maybe someone takes it up?

zoq:
> Looks good, just a couple comments about the build
> 
> * Does the style check job need to be updated in Jenkins? I think you have been working on these lately but I am not sure of the details.
> * I think for the Travis issue we shouldn't bother addressing it in this PR but I think the 'usual' MNIST download source can be unreliable.  It might be better to use mlpack.org/datasets; not sure though.

I haven't touched the style check job, but I know what the issue is; the plugin we use to parse the xml result now expects a correctly formatted xml. I can fix that later today.

movie_lens_prediction_with_cf/movie-lens-cf-cpp.ipynb
rcurtin: I see, yeah, maybe there is no better solution for now and opening an issue is a good idea.  I don't know of a header macro we can use to detect whether we're in a xeus-cling environment.  Maybe it would be a good idea to just add a comment like "this header is only necessary in a notebook environment" or something everywhere that it's included?

rcurtin:
No worries about the CI jobs, I was just confirming that this PR isn't breaking them.

zoq:
Let's squash this PR in the merge process.

rcurtin (approved):
Looks good to me; sorry for the slow approval. :+1:

PR #163, opened 2021-06-30
Use datasets.mlpack.org as the source for the datasets

rcurtin:
I unpacked `datasets.tar.gz` into `datasets.mlpack.org/`.  That should make it a bit easier to maintain the website, and simplify the website building scripts.  So, this PR makes all the download scripts use that subdomain as a source for the datasets. Let me know if I overlooked anything! :)


rcurtin:
(See also https://github.com/mlpack/mlpack.org/pull/51)

zoq (approved):
:+1:

PR #168, opened 2021-07-12
Install SoS kernel

zoq:
The package allows us to run different kernel in the same notebook, and also exchange data between different kernel. For example we can do run mlpack's logistic regression method with the C++ kernel, and plot the data with Python or create an interactive plot in JavaScript.

For some more details see https://vatlab.github.io/sos-docs/

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -51,3 +51,7 @@ dependencies:
 - gxx_linux-64
 - mlpack
 - ensmallen
+# SoS
+- sos-notebook
+- jupyterlab-sos
+- sos-papermill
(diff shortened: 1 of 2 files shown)


coatless (approved):

jonpsy:
Question: will this render all our utils along with matplotlib-cpp useless?

coatless:
Not at all. This just provides convenience for folks who are used to using Python visualization libraries. 

`matplotlib-cpp` shines because it handles the glue without needing a specialized kernel (SoS, in this case).

kartikdutt18 (approved):
LGTM!

jonpsy:
As the wise saying goes:

"LESGO!!"
           ~DaBaby.

PR #170, opened 2021-07-21
Added shell script for setting up local environment

Davidportlouis:
Shell script automates the task of installing all required dependencies required to replicate lab.mlpack.org environment in our local machine.

Supports:
- [x] - Creating new conda environment & installing dependencies.
- [x] - Upgrading dependencies in existing conda environment.

Usage:
```
$ wget -q https://raw.githubusercontent.com/davidportlouis/examples/env_setup/install.sh
$ chmod a+x install.sh
$ ./install.sh <environment name of your choice>
```


install.sh
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function installDeps() {
+    echo "*********************************"
+    echo "Upgrading environment using mamba"
+    echo "*********************************"
+    if [[ -f "./environment.yml" ]]; then
+        echo " "
+    else
+        wget -q https://raw.githubusercontent.com/mlpack/examples/master/binder/environment.yml
```
zoq, line 28: Does conda come with `wget` or does that use `wget` from the host system, maybe a good idea to install wget in the conda env?

Davidportlouis: In my case conda uses wget from host system, I think it maybe good idea to install wget in conda env in case host system does not have wget pre-installed.

Davidportlouis:
I did the environment setup using install.sh on my new machine.
Screenshot from 2021-07-22 21-53-05
It works great out of the box without any additional configuration, tweaks etc.

zoq (commented):
Nice, I'll test this on my machine next, btw. should we open another PR for the language specific modifications, or do you like to do it as part of this PR?

Davidportlouis:
> Nice, I'll test this on my machine next, btw. should we open another PR for the language specific modifications, or do you like to do it as part of this PR?

I think it would be better to open a new PR, because we have to create separate environment.yml for each languages containing their specific packages.

install.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+function cleanUp() {
```
zoq, line 3: Can you create a new e.g. `scripts` folder and put the script in there, also might be good to rename the script to `jupyter-conda-setup.sh`  or something like this.

install.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
```
zoq, line 1:
```suggestion
#!/usr/bin/env bash
```

To make it more robust.

zoq:
Just tested the script, works great.

zoq (approved):
Thanks for putting this together, super useful.

kartikdutt18 (approved):
Thanks for this contribution. LGTM!

zoq:
Thanks, this is super useful.

PR #172, opened 2021-07-25
GANs mnist

swaingotnochill:
@kartikdutt18 @zoq I have created this new pull request coz in the earlier branch a lot of stuff has been messed up. 
To do:
1. Document Model architecture( It's half done).
2. Notebook to create samples( but this can be done after we train the model).

 Please check the correctness of script, especially the Evaluate Function.
 @zoq If this training file is correct, can you please train the model? 
Right, the model is running for 657 cycles with 1 epoch per cycle. May be we can improve on this? Your thoughts.

P.S: Ignore the file path, and make file configurations, I will fix them later on.


mnist_gan/mnist_gan.cpp
```diff
@@ -0,0 +1,205 @@
...
+                  ens::PrintLoss(),
+                  ens::ProgressBar(),
+                  ens::Report());
+
+        optimizer.ResetPolicy() = false;
+        std::cout << " Model Performance " <<
+                  gan.Evaluate(gan.Parameters(), // Parameters of the network.
+                               i, // Index of current input.
+                               batchSize); // Batch size.
+    }
```
kartikdutt18, line 198: We don't need to use loops anymore, You can refer to [mnist_cnn](https://github.com/mlpack/examples/blob/4b93fecbf053d796bf3b0337e6f49a14ef17b335/mnist_cnn/mnist_cnn.cpp#L173).

swaingotnochill:
@zoq @kartikdutt18 I forgot to mention that the split function is not working for me. Like it doesn't do anything, and the entire dataset is loaded. Can you check if it is the same for you?

mnist_gan/mnist_gan.cpp
```diff
@@ -0,0 +1,205 @@
...
+                  optimizer,
+                  ens::PrintLoss(),
+                  ens::ProgressBar(),
+                  ens::Report());
+
+        optimizer.ResetPolicy() = false;
+        std::cout << " Model Performance " <<
+                  gan.Evaluate(gan.Parameters(), // Parameters of the network.
+                               i, // Index of current input.
+                               batchSize); // Batch size.
```
kartikdutt18, line 197: For saving periodically, you can use https://github.com/mlpack/models/blob/master/ensmallen_utils/periodic_save.hpp

zoq:
@swaingotnochill @Davidportlouis I just realized that we have a notebook that shows how you can load images into an armadillo matrix with the format mlpack expects - https://github.com/mlpack/examples/blob/master/cifar10_transformation_with_pca/cifar-10-pca-cpp.ipynb. Let me know if you have any questions about the notebook.

mnist_gan/mnist_gan.cpp
```diff
@@ -0,0 +1,205 @@
...
+    arma::mat trainDataset, valDataset;
+    data::Split(mnistDataset, trainDataset, valDataset, trainRatio);
+
+    std::cout << " Dataset Loaded " << std::endl;
+    std::cout << " Train Dataset Size : (" << trainDataset.n_rows << ", " << trainDataset.n_cols << ")" << std::endl;
+
+    std::cout << " Validation Dataset Size : (" << valDataset.n_rows << ", " << valDataset.n_cols << ")" << std::endl;
+
+    arma::mat trainTest, dump;
+    data::Split(trainDataset, dump, trainTest, 0.045);
```
zoq, line 64: If I check the output against the master build it looks good. Did you run this against the master branch as well or the latest stable release? If you run this against the latest stable release did you remove the header from the csv file?

swaingotnochill: I was using the master branch while compiling( we made those changes together if you remember). I haven't tested with the stable release version.

zoq: Can you post the output you get?

swaingotnochill: It's simple as the Validation dataset size is (0,0)

zoq: This is so strange, I'm using the exact same code and get:
```
785x42000
 Dataset Loaded
 Train Dataset Size : (784, 8400)
 Validation Dataset Size : (784, 33600)
```

zoq: Can you run:

`const size_t testSize = static_cast<size_t>(mnistDataset.n_cols * trainRatio);`

and drop the output here, that line is used inside the split function to set the size of the test matrix.

swaingotnochill:
`785x42000
 Dataset Loaded 
 Train Dataset Size : (784, 42000)
 Validation Dataset Size : (784, 0)
Training ...`
I am not sure why....I did build mlpack from master...maybe I will create a new build and test it...

mnist_gan/mnist_gan.cpp
swaingotnochill:
``` 785x42000
 Dataset Loaded 
 Train Dataset Size : (784, 42000)
 Validation Dataset Size : (784, 0)
 testSize : 0 
```

swaingotnochill: I am rebuilding from master ( just to make sure its not my build issue)...I think its already past 12 am midnight...you should get some sleep...let's talk back in morning...

zoq: Hm, can you also print `mnistDataset.n_cols` and `trainRatio`. The size of `mnistDataset` should not change after you call the split function, and I think you are still using `trainRatio = 0.8`?

zoq: `Train Dataset Size : (784, 42000)` --> `42000`  is also the entire dataset.

zoq: I can reproduce the output if I set `trainRatio = 0`, so I wanted to make sure `0.8` is what you use for your test as well.

zoq: Ohh, wait, `size_t trainRatio = 0.8;`, the data type is wrong, it should be `double`.

swaingotnochill:
```
After Split: 
 testSize0
 Mnist_Dataset cols:  42000
 TrainRatio: 0
```
Now this is interesting...after split trainRatio value becomes 0.

zoq: No, see the comment above.

mnist_gan/mnist_gan.cpp
```diff
@@ -0,0 +1,205 @@
...
+using namespace mlpack::data;
+using namespace mlpack::ann;
+using namespace mlpack::math;
+using namespace mlpack::regression;
+using namespace std::placeholders;
+
+
+int main()
+{
+    size_t trainRatio = 0.8;
```
zoq, line 22: I think that was solved the in the latest version, can you push the updated version?

swaingotnochill:
Sure...

mnist_gan/mnist_gan.cpp
swaingotnochill: I will update by 18:00utc

mnist_gan/mnist_gan_generate.cpp
```diff
@@ -0,0 +1,110 @@
...
+    std::function<double ()> noiseFunction = [](){ return math::Random(-8, 8) +
+    math::RandNormal(0, 1) * 0.01;};
+
+    FFN<SigmoidCrossEntropyError<> > generator;
+
+    FFN<SigmoidCrossEntropyError<> > discriminator;
+
+    GaussianInitialization gaussian(0,1);
+
+    GAN<FFN<SigmoidCrossEntropyError<> >, GaussianInitialization,
```
swaingotnochill, line 74: @zoq @kartikdutt18 Is it the correct way to use the Gan class. I declared generator, discriminator and gaussian above to use the Gan class. Let me know.

swaingotnochill:
The generate_image.py file is incomplete[I have yet to implement what kartik proposed]. Also the code needs a bit of cleaning. Please let me know if there is any logic error.

swaingotnochill:
@zoq @kartikdutt18 I am preparing the final notebook, but it seems that it crashes very often. Will it be alright if we can have the model definition, and the explanation part on the notebook.

utils/generateimage.hpp
```diff
@@ -0,0 +1,95 @@
...
+  pName = PyUnicode_DecodeFSDefault("generateimage");
+
+  pModule = PyImport_Import(pName);
+  Py_DECREF(pName);
+
+  if (pModule != NULL)
+  {
+    // The Python function from the generateimage.py script
+    // we like to call - cgenerateimage
+    pFunc = PyObject_GetAttrString(pModule, "cgenerateimage");
```
zoq, line 40: Does, this function fail right away, or does this actually generate the image and fail afterwards?

swaingotnochill: The Python Script works..but when I use the C++ Function, it throws "Call Failed".

swaingotnochill: Is it working on your system?

zoq: I just looked at the code, and haven't seen an issue, so I was wondering if the python script is called, generates an image but fails afterwards. Because in some cases we don't check for errors see https://github.com/mlpack/examples/blob/master/utils/plot.hpp

swaingotnochill: Yaa..not sure why it is throwing an issue...I think, even though the python script seems to be working fine alone, maybe the object("File Saved") is the root cause of this issue. I will try with different save image function, and see if this continues. I will update you by evening.

swaingotnochill: Its working now. Resolving this.

swaingotnochill:
@zoq @kartikdutt18 Can you review the code? See if this can merge. Later, after the model is trained, we can just push the updated output. As for the notebook, I have just added the output code. The model architecture is written in comments in the mnist_gan.cpp file. Just let me know if we want to write the same model architecture in notebook too. Also, if there is anything to add, feel free to tell.

mnist_gan/.gitignore
```diff
@@ -0,0 +1,7 @@
+.ipynb_checkpoints
```
zoq, line 1: All of this is covered in https://github.com/mlpack/examples/blob/master/.gitignore, so we can remove this one.

mnist_gan/.gitignore
```diff
@@ -0,0 +1,7 @@
...
+saved_models
+mnist_gan
+mnist_gan_generate
+mnist_gan_generate.o
+mnist_gan.o
+.vscode
```
kartikdutt18, line 7: Lets add an new line here (same for the file below).

mnist_gan/mnist_gan.cpp
```diff
@@ -0,0 +1,214 @@
...
+     * Note: Output of a Convolution layer = [(W-K+2P)/S + 1]
+     * where, W : Size of input volume
+     *        K : Kernel size
+     *        P : Padding
+     *        S : Stride
+     */
+
+
+  // Creating the Discriminator network.
+    FFN<SigmoidCrossEntropyError<> > discriminator;
```
kartikdutt18, line 106: The spacing is off here.

mnist_gan/mnist_gan.cpp
```diff
@@ -0,0 +1,214 @@
...
+    ens::Adam optimizer(stepSize, // Step size of optimizer.
+                        batchSize, // Batch size.
+                        0.9,       // Exponential decay rate for first moment estimates.
+                        0.999,     // Exponential decay rate for weighted norm estimates.
+                        eps,       // Value used to initialize the mean squared gradient parameter.
+                        numIterations,  // iterPerCycle// Maximum number of iterations.
+                        tolerance,    // Tolerance.
+                        shuffle);     // Shuffle.
+    std::function<double()> noiseFunction = [] () {
+        return math::RandNormal(0, 1);};
```
kartikdutt18, line 182:
```suggestion
    std::function<double()> noiseFunction = []() {
        return math::RandNormal(0, 1);};
```

mnist_gan/mnist_gan_generate.cpp
```diff
@@ -0,0 +1,116 @@
...
+  //   inputData = (inputData - 0.5) * 2;
+
+  //   data::Split(inputData, trainData, validData, 0.8);
+  // }
+
+    arma::arma_rng::set_seed_random();
+
+    // Define noise function.
+    std::function<double ()> noiseFunction = [](){ return math::Random(-8, 8) +
+    math::RandNormal(0, 1) * 0.01;};
```
kartikdutt18, line 53: Indentation seems off here.

mnist_gan/mnist_gan_generate.cpp
```diff
@@ -0,0 +1,116 @@
...
+    // Matrix to store the generated data.
+    arma::mat generatedData(2 * dim, dim * numSamples);
+
+
+    for (size_t i = 0; i < numSamples; ++i)
+    {
+    arma::mat samples;
+
+    // Create random noise using noise function.
+    noise.imbue( [&]() { return noiseFunction(); } );
```
kartikdutt18, line 92:
```suggestion
    noise.imbue([&]() { return noiseFunction(); });
```

mnist_gan/mnist_gan_generate.cpp
```diff
@@ -0,0 +1,116 @@
...
+    samples = samples.t();
+    generatedData.submat(dim,
+        i * dim, 2 * dim - 1, i * dim + dim - 1) = samples;
+    }
+    // Save the output as csv.
+    data::Save("./samples_csv_files/sample.csv", generatedData, false, false);
+
+    std::cout << "Output generated!" << std::endl;
+
+}
```
kartikdutt18, line 116: We can add a new line here.

kartikdutt18 (commented):
Some minor changes.

swaingotnochill:
> @swaingotnochill Is there anything else required for this PR? This addition is very nice.

The saved model is only trained for 2 epochs...We can merge this, just that we have to train the model for this which might take some days. But, the architecture seems fine for now, so shouldn't be an issue if we add this..

shrit:
@zoq do we merge this?

shrit:
I will try to port the C++ example manually to the examples repo.

shrit:
@geekypathak21  feel free when you have the time to port this one as well, many thanks

geekypathak21:
@shrit  Sure will do :+1:

PR #174, opened 2021-07-30
Test the travis build...

rcurtin:
Don't merge this---I just want to make sure the transition from travis-ci.org to travis-ci.com was successful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-blah blah blah
+blah blah
 
 The mlpack **examples** repository contains simple example usages of mlpack.
 You can take the code here and adapt it into your application, or compile it and
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+blah blah blah
+
 The mlpack **examples** repository contains simple example usages of mlpack.
 You can take the code here and adapt it into your application, or compile it and
 see what it does and play with it.


zoq:
I had the same problem with ensmallen, I thought maybe just transition to azure. I can set this up.

rcurtin:
Yeah, I am looking through their pricing model and it is a nightmare.  I think I have us set up with a "free" 10k credit plan, but presumably it will run out soon enough.  If you want to transition to Azure---or Jenkins---that would be a way better solution.

zoq:
Yeah, Jenkins would probably be even easier to set up.

rcurtin:
Ok, that at least triggered.  Apparently 10k credits is roughly 1000 minutes of build time... since each build takes roughly half an hour, that gives us... just a couple more than 30 builds before we're out.  Sigh...

I think Jenkins might be the better way to go here, but we should keep the build script in this repository instead of in the Jenkins config, and the Jenkins job can just run that script.

zoq:
Agreed, we wanted to set up testing for the notebooks as well, so that would fit perfectly.

rcurtin:
Hmm, I guess we could do the builds in a docker container too to save the extra setup time of installing mlpack's dependencies.  (I guess we still have to build and install mlpack though.)

zoq:
If we do this inside a docker conda env we could install the nightly ensmallen/mlpack conda package, instead of building mlpack each time.

rcurtin:
That works too---do you want to set that up?  If not I'll put it on my list (could be a few days).  Anyway at least this is working so I'll close this PR. :+1:

PR #179, opened 2021-08-20
MNIST noisy image reconstuction using AutoEncoder

Davidportlouis:
- [x] Added PyTorch model and model weights
- [x] Added weight_convertor for extracting weights from PyTorch model 
- [x] Load PyTorch model weights into mlpack model.
- [ ] Add noisy to MNIST test set for evaluation (adding noisy externally is not working).
- [ ] Notebook for visualization the reconstructed images (not performing as expected).
- [ ] Dataset Source.


mnist_denoising_AutoEncoder/denoiseNet_load_weights.cpp
```diff
@@ -0,0 +1,241 @@
...
+  model.Add<TransposedConvolution<>>(16, 32, 2, 2, 2, 2, 0, 0, 7, 7, 14, 14);
+  model.Add<LeakyReLU<>>(0.1);
+  
+  // layer_8
+  model.Add<TransposedConvolution<>>(32, 64, 2, 2, 2, 2, 0, 0, 14, 14, 28, 28);
+  model.Add<LeakyReLU<>>(0.1);
+
+  // classifier
+  model.Add<Convolution<>>(64, 1, 3, 3, 1, 1, 1, 1, 28, 28);
+  //model.Add<Sigmoid<>>(); //NEW_Q: Can I use Sigmoid(), I don't see a Sigmoid Implementation in ann namespace, my python model using sigmoid in the output layer ?
```
kartikdutt18, line 201: You can use logistic function [https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/activation_functions/logistic_function.hpp] here.

mnist_denoising_AutoEncoder/denoiseNet_load_weights.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+
+  model.ResetParameters();
+  model.Parameters().fill(0.0);
+  std::cout << model.Parameters().n_elem << std::endl;
+  LoadWeights<>(model, "./cfg/denoiseNet.xml");
+  std::cout << model.Parameters().n_elem << std::endl;
+  mlpack::data::Save("ae.xml", "model", model, false);
+  // Can I comment the below line ?
```
kartikdutt18, line 210: Yes

mnist_denoising_AutoEncoder/models/denoiseNet/denoiseNet.py
```diff
@@ -0,0 +1,80 @@
+from collections import OrderedDict
```
zoq, line 1: Might be better to move this to https://github.com/mlpack/models, @Davidportlouis @kartikdutt18 what do you think? And use that inside the notebook itself.

Davidportlouis: Sure

mnist_denoising_AutoEncoder/denoiseNet_load_weights.cpp
Davidportlouis: Any examples or tests related to usage of logistic loss would be helpful

mnist_denoising_AutoEncoder/denoiseNet_load_weights.cpp
Davidportlouis: Ok

mnist_denoising_AutoEncoder/denoiseNet_load_weights.cpp
kartikdutt18: Not sure if this would be helpful, https://github.com/mlpack/mlpack/blob/499f9d81d63d508e48d7a2cdea35bc427def2087/src/mlpack/tests/activation_functions_test.cpp#L773 . I can share a code snippet if you want

Davidportlouis: I took a look at it, was unable to get exactly how to use it. Code snippet would be great

kartikdutt18: You can try : 
```
FFNObject->Add<SigmoidLayer<>>();
```
Reference : https://github.com/mlpack/mlpack/blob/499f9d81d63d508e48d7a2cdea35bc427def2087/src/mlpack/methods/ann/layer/base_layer.hpp#L146

Davidportlouis: I think this might work, we need to find a way to add random noisy to mnist_test data. The method I tried adding random noisy and clliping using NumPy externally is completely destroying the images.

PR #180, opened 2021-08-21
Add imbalanced-learn conda pkg to lab.mlpack environment

Davidportlouis:
Imbalanced-learn (imblearn) is required by SMOTE technique used in our resampling utility for c++ notebooks.


zoq (approved):
Thanks!

PR #182, opened 2021-09-26
Trigger a new binder build.

zoq:
This is just to force a new binder build, so we get the latest sos-notebook package which fixes the await bug.


rcurtin (approved):
Looks good; not sure if you need to merge this, or just needed to trigger the build, but it seems fine to me either way. :+1:

zoq:
This needs to be merged, binder caches the images, so to build a new images, we have to update the repo.

PR #190, opened 2022-02-20
Fix binder "Error during build: .0.0 is not valid SemVer string" build issue.

zoq:
Fix binder "Error during build: .0.0 is not valid SemVer string" build issue.

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -16,7 +16,7 @@ dependencies:
 - altair=4.1
 - bqplot=0.12.20
 - dask=2020.12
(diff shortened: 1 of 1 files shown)


rcurtin (approved):
I'm not sure I understand why this fixes the problem, but if it fixes it, I don't need to ask too many questions especially with such a simple diff. :)

kartikdutt18 (approved):

rcurtin:
I verified that this works!  Now I understand a little more about the notebook setup process. :)

PR #192, opened 2022-03-09
Switched to Internal split for LSTM examples

### LINKED ISSUE #79: Switch to Internal split for LSTM examples.
LSTM examples requires data that is not shuffled. Currently mlpack's internal `data split` class doesn't allow us to split data without shuffling yet. After mlpack/mlpack#2293 is merged we should be able to do that. This issue aims to make the change from the current implementation in LSTM examples to internal split.
Let me know if I need to clarify anything.
Thanks a lot.

eshaanagarwal:
Fixes #79 using https://github.com/mlpack/mlpack/pull/2293

### DIFF
--- a/lstm_stock_prediction/lstm_multivariate_time_series_prediction.ipynb
+++ b/lstm_stock_prediction/lstm_multivariate_time_series_prediction.ipynb
@@ -200,10 +200,9 @@
     "size_t inputSize = 5, outputSize = 2;\n",
     "\n",
     "// Split the dataset into training and validation sets.\n",
-    "arma::mat trainData = dataset.submat(arma::span(),arma::span(0, (1 - RATIO) *\n",
-    "      dataset.n_cols));\n",
-    "arma::mat testData = dataset.submat(arma::span(), arma::span((1 - RATIO) * dataset.n_cols,\n",
-    "      dataset.n_cols - 1));"
+    "arma::mat trainData;\n",
+    "arma::mat testData;\n",
+    "data::Split(dataset, trainData , testData, RATIO, false);"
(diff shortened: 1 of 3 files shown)


lstm_stock_prediction/lstm_multivariate_time_series_prediction.ipynb
```diff
@@ -200,10 +200,9 @@
     "size_t inputSize = 5, outputSize = 2;\n",
     "\n",
     "// Split the dataset into training and validation sets.\n",
-    "arma::mat trainData = dataset.submat(arma::span(),arma::span(0, (1 - RATIO) *\n",
-    "      dataset.n_cols));\n",
-    "arma::mat testData = dataset.submat(arma::span(), arma::span((1 - RATIO) * dataset.n_cols,\n",
-    "      dataset.n_cols - 1));"
+    "arma::mat trainData;\n",
+    "arma::mat testData;\n",
+    "data::Split(dataset, trainData , testData, RATIO, false);"
```
rcurtin, line 205:
```suggestion
    "data::Split(dataset, trainData, testData, RATIO, false);"
```

rcurtin (commented):
Awesome, thanks!  Did you check that the code still achieves the same performance before and after the change?  (I am in the process of doing that, but I think #190 has to be solved first before I can run these.)

rcurtin (approved):
Sorry it took so long to get back to this!  Everything looks good to me.  Thanks for the contribution. :)

rcurtin:
Oops! Forgot to hit merge. Thanks again @eshaanagarwal!

PR #193, opened 2022-03-10
Bump to latest versions

dnabanita7:


rcurtin:
What does this solve and why should we merge it?  There's no description here, so I don't understand what the PR is for.  (If there was already discussion somewhere else and I just don't have context, oops, sorry about that.)

zoq:
Probably a good idea to keep the packages updated, did you test all notebooks, to make sure they are still working?

dnabanita7:
Yes, my bad! Sorry for not describing it properly. The previous versions of matplotlib and soem other packages are not suitable for the python version so had to upgrade it.

rcurtin:
Why are they not suitable?

dnabanita7:
Matplotlib3.1 doesn't support py 3.7 anymore.

zoq (dismissed):
Thanks for looking into this.

zoq:
Hm, looks like we run into another package version issue.

dnabanita7:
I will check this again

zoq:
Thanks, what I did was just opened the binder link from the second post, it worked before, but looks like since then there was another package update.

dnabanita7:
works fine now!

zoq (approved):
Checked, builds and works, thanks for the updates.

zoq:
Thanks for the update.

PR #195, opened 2022-03-21
New classification example

tareknaser360:
Added a new example notebook using several classification algorithms


iris-classification/Iris.csv
```diff
@@ -0,0 +1,151 @@
+Id,SepalLengthCm,SepalWidthCm,PetalLengthCm,PetalWidthCm,Species
```
zoq, line 1: We usually, don't push datasets to the repo, to keep it small. I'll upload the data to the mlpack page so we can fetch it from there.

zoq (commented):
Can you remove the html file as well, would be nice to eport each notebook as an html page and put it on the website, but we shouldn't put it in the repo.

iris-classification/Iris.csv
tareknaser360: Ok I will update the notebook.
Would to upload the dataset to datasets.mlpack.org ?

PR #201, opened 2022-07-26
Added folium

tareknaser360:


zoq (approved):
Thanks for putting this together.

PR #203, opened 2022-08-14
Mnist CNN Makefile

shubham1206agra:
Updating version of c++ to c++14

PR #204, opened 2022-08-16
Update all Makefiles to use C++14.

rcurtin:
This is a simple follow-up to #203 that just applies the same changes to the rest of the Makefiles.


kartikdutt18 (approved):

zoq (approved):
I can fix the build, unless you already started working on it.

rcurtin:
> I can fix the build, unless you already started working on it.

Sure, feel free---i haven't started looking into it, but I think it should be pretty straightforward to fix.

PR #207, opened 2022-10-12
Force version of libstdc++ to 9.3.0 to fix cling kernel

rcurtin:
It seems like somehow libstdc++ got upgraded to 9.5.0 in the conda environment, but this is incompatible with the version of cling that gets installed.

This fix just forces the version to be 9.3.0, so it works again.

I also removed `ijavascript` because keeping it in there caused some build issue that I did not take the time to debug.

Also, I found that the system uses Python 3.7, not 3.9, so I updated the path in xeus-cling.hpp.

I tested this on mybinder.org and ran through the Avocado price prediction notebook successfully.

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -51,7 +51,7 @@ dependencies:
 - cython
 - numpy
 - pandas
-- gxx_linux-64
+- gxx_linux-64=9.3.0
 - mlpack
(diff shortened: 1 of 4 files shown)


zoq (approved):
Nice.

PR #208, opened 2022-10-14
Adapt notebooks to mlpack 4 API

rcurtin:
I went through all of the C++ notebooks and made changes that reflect the latest changes in mlpack master (which will be mlpack 4).  Basically these changes are:

 * Include only `<mlpack.hpp>`. (mlpack/mlpack#3265)
 * Remove all superfluous namespace declarations (mlpack/mlpack#3269)
 * Moved datasets to https://datasets.mlpack.org/ for consistency.

I also made a few fixes where necessary, and sometimes I reduced the computational complexity of the models we were learning so we would stay within resource limits.

Now, every notebook runs successfully on Binder, with two exceptions:

 * The reinforcement learning notebooks.  I didn't touch those.
 * The `cifar10_eval` notebook needs an 8GB binder instance to succeed.  I had one of those once, when starting a notebook on Binder, but then I started getting 2GB instances.  However, it should work.

After merging this PR, https://datasets.mlpack.org/cifarNet2.xml will need to be moved to https://datasets.mlpack.org/cifarNet.xml.

There is one additional issue: datasets.mlpack.org is not available from OVH binder instances.  I'm trying to work with the Binder Gitter chat and Github repository to get a resolution there.


avocado_price_prediction_with_linear_regression/avocado_price_prediction_with_lr_cpp.ipynb
```diff
@@ -512,19 +510,27 @@
...
+    "To perform linear regression we'll be using the `LinearRegression` class from mlpack."
    ]
   },
   {
    "cell_type": "code",
    "execution_count": 22,
    "id": "97fdc354-57d0-4e28-91e8-1d707fc24226",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
zoq, line 521: I guess, we got lucky before.

rcurtin: Yeah.  I suppose we could increase the penalty to work around the issue, but I don't think it's a big deal either way.

zoq: The results are pretty much the same, so I don't think this is needed.

zoq (approved):
Looks good to me, no comments from my side.

PR #209, opened 2023-02-22
Minor Syntax Error

AdarshSantoria:


rcurtin (commented):
Thanks for the fix!  I believe that the only layer that still needs to be adapted for this example to work is the `PReLU` layer, which is being handled in mlpack/mlpack#3420.  Once that's merged, do you want to uncomment the code that skips the `mnist_cnn` build in `.ci/linux-steps.yaml` and `.ci/macos-steps.yaml` and see if the example builds and runs correctly?

AdarshSantoria:
keep open

rcurtin:
Sorry that mlpack-bot got to this one before I did. :)  Let's see if it builds, and if it's green, then I think we can merge it.  Thanks for working on this! :+1:

AdarshSantoria:
Hey @rcurtin, I am getting this in binder
```
bash: line 1:     9 Killed                  ${MAMBA_EXE} env update -p ${NB_PYTHON_PREFIX} --file "binder/environment.yml"
time: 1009.813
```
Can you please help or just provide me some paths to solve this.

rcurtin:
Oops, sorry, this one fell off the list.  Let me try restarting all the checks and let's see what happens, and we can go from there.  Definitely it would be great to merge this fix and get these examples working again.

rcurtin (approved):
@AdarshSantoria I'm not sure what the error you were encountering was (looks like notebook worker out of RAM?), but after I re-enabled the `mnist_batch_norm` build in its `Makefile`, everything works in the example, so I think this is good to merge. :+1:  Thanks for the fix!

AdarshSantoria:
Thanks for the help.

zoq (approved):
Looks good to me.

AdarshSantoria:
Keep open

rcurtin:
Oops, forgot to actually hit the merge button...

PR #213, opened 2023-05-15
Apply the same preprocessing to the test set as to the training set

rcurtin:
I noticed while reading the CI logs for #209 that the test accuracy on `mnist_batch_norm` and `mnist_simple` was atrocious (9%!) even though the training and validation accuracies seemed fine.  A quick look indicated that this was because we were using the wrong dimension for labels, and for `mnist_simple` we weren't normalizing the test data.  Easy fix, and now the test accuracies for both are what is expected (95%+).

### DIFF
--- a/mnist_batch_norm/mnist_batch_norm.cpp
+++ b/mnist_batch_norm/mnist_batch_norm.cpp
@@ -172,8 +172,8 @@ int main()
   // https://www.kaggle.com/c/digit-recognizer/data
 
   data::Load("../data/mnist_test.csv", dataset, true);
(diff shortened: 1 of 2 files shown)


birm (approved):
I had some issues checking this out with the binder link (seemed to freeze at step 40/63 when looking for the python dependencies; maybe it was just slow, but I came back some time later where it had just nondescriptly errored. I'll try it again next PR).

That said, this clearly fixes the issue. Hope you're doing well rcurtin!

rcurtin:
Thanks!  Yeah, doing well, just chugging away at a first release of bandicoot. :)  Hope things are well for you too!

PR #214, opened 2023-08-14
Reinforcement Learning: Examples for `DDPG` and `TD3` with Gymnasium Environments

tareknaser:
## Description:
This PR updates the reinforcement learning examples with the latest Gymnasium Python server. Two significant examples have been added.
### `DDPG` on `MountainCarContinuous-v0`:
- The agent was trained for a total of 100,000 steps.
### `TD3` on `Pendulum-v1`:
- The training was performed for 50,000 steps.

> Note: Both examples include locally saved videos that will be utilized in my final GSoC report.


reinforcement_learning_gym/mountain_car_ddpg/server.py
```diff
@@ -0,0 +1,477 @@
...
+          compressionLevel = int(compression)
+        except ValueError:
+          compressionLevel = 0
+          close = True
+
+      actionspace = get_optional_params(jsonMessage, "env", "actionspace")
+      if isinstance(actionspace, basestring):
+        if actionspace == "sample":
+          sample = env.action_space.sample()
+          # sample = envs.get_action_space_sample(instance_id)
```
zoq, line 369: I think, we should remove the server, and make the gym tcp a dependency.

reinforcement_learning_gym/pendulum_td3/server.py
```diff
@@ -0,0 +1,477 @@
+import json
```
zoq, line 1: See the comment above.

reinforcement_learning_gym/mountain_car_ddpg/server.py
tareknaser: Yes I think this one slipped
I was using it from the same folder for convenience and pushed it

zoq (approved):

zoq:
Thanks for the grate contributions.

PR #191, opened 2022-03-02
Restructuring examples folder

dnabanita7:
- [x] Create an examples sub-folder
- [x] Create respective language sub-folders and add their respective README.md
- [x] Add links in the main README.md for easy access
- [x] Describe the changes in main README.md as well as in respective example README.md
- [x] Check if nothing breaks due to change of structure

Ref: #154

PR #215, opened 2023-09-20
Adding float mnist_simple example

shrit:


mnist_simple_float/mnist_simple.cpp
```diff
@@ -0,0 +1,184 @@
...
+              //ens::EarlyStopAtMinLoss(
+                  //[&](const arma::fmat& [> param <])
+                  //{
+                    //double validationLoss = model.Evaluate(validX, validY);
+                    //cout << "Validation loss: " << validationLoss << "."
+                        //<< endl;
+                    //return validationLoss;
+                  //}),
+              //// Store best coordinates (neural network weights)
+              //bestCoordinates);
```
shrit, line 138: I forgot to do this, I need to check ensmallen.

Do not review I will ping you when this is finished

conradsnicta:
@shrit Apologies - I don't have the bandwidth for this at the moment.  Suggest you try @zoq in addition to @rcurtin for the review.

mnist_cnn_float/mnist_cnn.cpp
```diff
@@ -0,0 +1,229 @@
...
+  // Calculate accuracy on validation data points.
+  double validAccuracy =
+      accu(predLabels == validY) / (double) validY.n_elem * 100;
+
+  cout << "Accuracy: train = " << trainAccuracy << "%,"
+            << "\t valid = " << validAccuracy << "%" << endl;
+
+  data::Save("model.bin", "model", model, false);
+
+  cout << "Predicting on test set..." << endl;
```
rcurtin, line 206: I looked through the test output and it seems like something fails here after this output:
https://dev.azure.com/mlpack/84320e87-76e3-4b6e-8b6e-3adaf6b36eed/_apis/build/builds/9352/logs/15

```
2023-09-21T12:51:20.9728229Z Predicting on test set...
2023-09-21T12:51:20.9728383Z Example failed: mnist_cnn_float
```

But, unfortunately, the CI output isn't more than that.  Does the whole example run for you successfully locally?

mnist_cnn_float/mnist_cnn.cpp
```diff
@@ -0,0 +1,229 @@
+/**
+ * An example of using Convolutional Neural Network (CNN) for
+ * solving Digit Recognizer problem from Kaggle website.
```
rcurtin, line 3: It's probably worth adding a comment pointing out that this is similar to the other example, but in this one we use fp32 precision.  (Same for the `mnist_simple` example.)

mnist_cnn_float/Makefile
```diff
@@ -0,0 +1,43 @@
+# This is a simple Makefile used to build the example source code.
+# This example might requires some modifications in order to work correctly on
+# your system.
+# If you're not using the Armadillo wrapper, replace `armadillo` with linker commands
+# for the BLAS and LAPACK libraries that you are using.
+
+TARGET := mnist_cnn
+SRC := mnist_cnn.cpp
```
rcurtin, line 8: Do you think we should rename the file `mnist_cnn_float.cpp` or similar to indicate that it is different from the other version?

rcurtin (commented):
Oh, I never saw this because I still did not have mailman3 working right for the mlpack-git list.  Sorry about that... I would have reviewed this earlier otherwise.  This is definitely a nice thing to include, but I wonder if calling the example `mnist_cnn_fp32` might be more clear for people scrolling through.  Alternately, I wonder if just making another file `mnist_cnn_fp32.cpp` in the `mnist_cnn` directory would suffice.  I don't think I have an opinion either way.

Fingers crossed that mailman3 successfully delivers this message to the mlpack-git list... :smile:

mnist_cnn_float/Makefile
shrit: will do the suggested above

shrit:
@rcurtin both of the examples has start training as usual, no problem while loading and training the data.
I will post more more if there is any issue related to prediction, but for me it seems that everything is running alright

shrit:
okay I am getting a segmentation fault during the prediction of the mnist_simple,  the reason for this is that I forgot to register the layers for float so the model matrix was empty and therefore armadillo throw the error.
I fixed this it should be fine now

rcurtin (approved):
Awesome, nice to add this.  I think we should probably reorganize all the examples at some point (like #191 or similar), but that doesn't change that we definitely should merge this.  Thanks for working out all the issues :+1:

PR #219, opened 2024-01-26
Bandicoot library example to train an mnist example

shrit:
Hi,

This is a draft and still does not compile.

### DIFF
--- a/mnist_simple_coot/mnist_simple_coot.cpp
+++ b/mnist_simple_coot/mnist_simple_coot.cpp
@@ -14,8 +14,10 @@
  * @author Omar Shrit
  */
 #define MLPACK_ENABLE_ANN_SERIALIZATION
-#include <mlpack.hpp>
+#define MLPACK_HAS_COOT
+
 #include <bandicoot>
+#include <mlpack.hpp>
 
 #if ((ENS_VERSION_MAJOR < 2) || \
     ((ENS_VERSION_MAJOR == 2) && (ENS_VERSION_MINOR < 13)))
@@ -30,7 +32,7 @@ coot::Row<size_t> getLabels(coot::mat predOut)
   coot::Row<size_t> predLabels(predOut.n_cols);
   for (coot::uword i = 0; i < predOut.n_cols; ++i)
   {
-    predLabels(i) = predOut.col(i).index_max();
+    // predLabels(i) = predOut.col(i).index_max();
   }
   return predLabels;
 }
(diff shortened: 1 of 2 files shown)


rcurtin (commented):
I'm looking forward to when this works!  We will also need to think about ensuring that the CI workers have a GPU so that we can actually test this example.  Or, something along those lines.

mnist_simple_coot/mnist_simple_coot.cpp
```diff
@@ -0,0 +1,188 @@
...
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ *
+ * @author Eugene Freyman
+ * @author Omar Shrit
+ */
+#define MLPACK_ENABLE_ANN_SERIALIZATION
+#define MLPACK_HAS_COOT
```
rcurtin, line 17: Just a TODO, we should integrate definition of `MLPACK_HAS_COOT` into the mlpack headers (maybe in `prereqs.hpp` or something, we try to detect Bandicoot), so that users don't have to write it by hand.  You're probably already thinking that, I just wanted to write it down so it doesn't get forgotten.

mnist_simple_coot/mnist_simple_coot.cpp
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ *
+ * @author Eugene Freyman
+ * @author Omar Shrit
+ */
+#define MLPACK_ENABLE_ANN_SERIALIZATION
```
rcurtin, line 16: Serialization will be a little tricky.  When this macro is enabled, we compile all the `serialize()` functions for all the layer types---for the type `arma::mat`.  Instead, we'll need to do `CEREAL_REGISTER_MLPACK_LAYERS(coot::mat)` or similar.  Also, we'll need to actually implement a `serialize()` function for Bandicoot types, a bit like in `src/mlpack/core/arma_extend/serialize_armadillo.hpp`.  I think the strategy should just be to convert it to an Armadillo matrix (i.e. pull it off the GPU into the CPU), and then serialize that, plus a little bit of code to put the Armadillo matrix back onto the GPU during loading.  Anyway, I can help with that when we get there.

mnist_simple_coot/mnist_simple_coot.cpp
shrit: Yes, on my list,

mnist_simple_coot/mnist_simple_coot.cpp
shrit: Yes, for now, the easy solution for now would be to convert it back to Armadillo.
In all cases, we are loading an armadillo matrix, so we will have to do the conversion at least once.
Once we have the load functionality for Bandicoot, if it makes sense, then we can think of serialization.

PR #222, opened 2024-03-12
df.corr should add additional argument to handle string data

tunglinwood:
1. (Minor) add `!pip install packages` including `statsmodels` , `folium`, and most important `mlpack` to avoid errors if user try on Jupyter notebook online.
2. (Major)since pandas version 2.0.0 now you need to add numeric_only=True param to avoid the issue


rcurtin:
Hey @tunglinwood, I did not realize that both this PR and #224 were open for the same thing.  I think that #224 is a more complete solution as it also fixes the other issues you pointed out in https://github.com/mlpack/mlpack/issues/3655, but the fix for `df.corr()` here is cleaner in my opinion.

For the `!pip install` bit, I wouldn't mind hearing @zoq's opinions and others who have been more involved with the development of Python notebooks in the past.  If we do it here, we should probably do it for all Python notebooks.  But I think that the intention was that these could be run in mlpack Binder images, so mlpack and other dependencies would already be installed.

PR #225, opened 2024-04-08
Refactor rl examples

Ali-Hossam:
This PR offers a refactored Reinforcement learning .cpp examples so that they work with the new mlpack API


rcurtin:
Are these just the C++ fixes from #223?  (If so we can just handle these changes there.)

Ali-Hossam:
Yes they are the same, i just noticed that i made two PRs instead of one. Should i close #223 and reopen this? @rcurtin

rcurtin:
Nah, #223 is just fine :+1:

PR #230, opened 2024-06-27
Example embedded

shrit:
I have added one example for now, and a set of CMakefile that I have used over the years to compile mlpack and integrate it into any application. I have never seen any problem with these and they always worked as they should.

The CMakeLists.txt is a simplified version of mlpack original CMakeList.txt and removed everything that is not relevant for a standing application.

PS: it might look like one directory is a duplicate of the other, but this is intended, the first one is a fully working example with a random forest, and the second one is just a set of CMake files that can be copied or adapted, and they are standalone.


rcurtin (commented):
Here are some high-level comments---I haven't gone through at a low level of detail yet.

Is CMake the right way to provide this example?  All of the other examples have just a Makefile that's very simple (by design).  I get that the functionality here is more complicated; there is the autodownloader and the ability to set flags for a specific board, but it feels different than other examples.

I thought while I was writing this that what the example really is is a CMake template for writing embedded applications.  Do you think it makes more sense to simply make it into its own repository?  It would be possible to even modify it so the autodownloader downloads mlpack itself too, since it's header-only.

The other big question to me is, how do we point users towards this?  Should we add to the mlpack main documentation to add a link somewhere?  (If so, where?)

And then the last question, which I think is less important, is---the main.cpp currently is just the random forest quickstart or some other trivial example.  Should we work that into a more "typical" embedded application?  Maybe you're planning to do that later?

I guess my biggest thought has to do with organization: I just want to make sure that a user coming in with no context is able to find this example, adapt it to their needs, and *also* not get confused by the other examples or types of examples in the repository.  Maybe one idea is to split into three example repositories (simple Makefile applications, embedded applications with CMake infrastructure, notebooks)?

embedded/mlpack_embedded_cmake/README.md
```diff
@@ -0,0 +1,18 @@
+## CMake files for mlpack
+
+This is a set of CMakefiles that autodowload all the mlpack dependencies. The
+objective of this example is to provide the minimum configurations that allow
+to build an mlpack application with in the objective of embedded system.
```
rcurtin, line 5:
```suggestion
This directory contains a minimal and simple CMake configuration that
autodownloads all the mlpack dependencies.  This example can be used as a simple
starting point for building an mlpack application that will be compiled to an
embedded system.
```

Totally up to you, just a minor writing cleanup.

embedded/mlpack_embedded_cmake/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+These files contains the necessary cmake files that are required in the
+embedded context, especially the cross compilation step, and to build a
+statically linked binary, you can use these configs to build any mlpack
+application for any platform as long as you modify the configs accordingly.
```
rcurtin, line 10:
```suggestion
This example contains the necessary CMake files that are required in the
embedded context, especially for cross-compilation, and to build a statically
linked binary.  You can use this configuration to build any mlpack
application for any platform, as long as you modify the configuration
accordingly.
```
Only trivial grammatical fixes here

embedded/mlpack_embedded_cmake/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The main CMakeLists.txt assumes that the project name is `main` and that you
+have a file that is called `main.cpp` that contains the mlpack C++ code you are
+trying to compile. Feel free to modify these the way it suits you.
+
+If you are looking to integrate these CMakeFiles into your project, feel free
+to modify them the way you want. The point of these files to provide a simple
+working example and to keep things simple
```
rcurtin, line 18: It feels like this may already be captured by the previous paragraphs.  I think it's not necessary to keep this paragraph, but it doesn't matter much either way.

embedded/mlpack_embedded_cmake/README.md
```diff
@@ -0,0 +1,18 @@
+## CMake files for mlpack
```
rcurtin, line 1: Do you think it might be worthwhile to write a little bit about buildroot or how you might set up a cross-compilation environment here?  I guess it will be documented elsewhere.

rcurtin: It could also be nice to add a list of the CMake options that can be specified by this basic configuration (like `BOARD_NAME`, that's probably the most important one).

embedded/crosscompile_random_forest/main.cpp
```diff
@@ -0,0 +1,35 @@
+#include <mlpack.hpp>
```
rcurtin, line 1: A simple header comment could be helpful here, just indicating that the file is a super-simple random forest example that doesn't do anything useful, but can be adapted to some more complicated embedded application.

rcurtin (approved):
I think I understand the objective here a little bit better now after a discussion and looking more closely.  It's definitely a useful and important thing to add.  Some points that we talked about on the call (just for the sake of recording them somewhere):

 * It probably would be a good idea to rename `mlpack_embedded_cmake` or something like `cmake_template` or similar, to indicate that it's a template that can be used to start an embedded application.
 
 * Pointing people towards this example in the right way is probably the most important part here; we talked about an idea of updating `doc/index.html` or even adding an `examples.md` page or something like this that points to individual examples (including this one) that is directly accessible from the homepage.  Definitely that is not something to do here, just something to keep in mind in the background as we move forward.

 * I also think we may have to do some later restructuring of this repository to handle the three different "forms" of examples we have, but, I don't think either of us have a great idea for how that should look now.  As time goes on a better way will probably become clear :smile:

shrit:
I will need to push a couple of things to this PR and then we can merge it

embedded/mlpack_embedded_cmake/README.md
shrit: I am planning to add all of that to our documentation
For here, I will just add a usage example in the README

embedded/crosscompile_random_forest/main.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+using namespace mlpack;
+
+int main(int argc, char** argv)
+{
+  arma::mat dataset;
+  data::Load("../../data/covertype-small.csv", dataset);
+
+  // Labels are the last row.\n",
```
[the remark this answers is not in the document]
shrit: from Jupyter :face_with_diagonal_mouth:

Issue #235, opened 2024-09-04
Off by one matrix multiplication


CharMyFloat:
In the example located at 
cpp/neural_networks/mnist_simple the function at line 166 in the cpp file is:
dataset.shed_row(0); // Strip labels before predicting.
This function seems to cause an off by one error for the matrix multiplication. Without it I don't have any issues, the line before it ignores the first line that has the data labels I believe. Thanks for putting this together, great work.

rcurtin:
Thanks for the report!  Are you sure you have the right dataset?  I ran it locally and had no issues.

```
$ md5sum mnist_train.csv
5b49cf1b57fb9d6102b559d59d99df7c  mnist_train.csv
$ md5sum mnist_test.csv
c807df8d6d804ab2647fc15c3d40f543  mnist_test.csv
```

The data should be downloaded with the `download_data_set.py` script in the `scripts/` directory (see the bottom of the README).

(While reproducing, I noticed that the paths of test files were wrong, so I opened #236 to fix that.)

CharMyFloat:
Ah maybe that's it, I got it from kaggle. Out of curiosity what kind of Accuracy test % are you getting from this sample?

rcurtin:
I get this:

```
Validation loss: 976.893.
Accuracy: train = 98.8926%,      valid = 96.9162%
Predicting on test set...
Accuracy: test = 97.32%
```

But, worth keeping in mind that this is a truly trivial network; it's not too hard to build a more complex network that performs much better.  The `mnist_cnn` example will do a lot better, for instance.

CharMyFloat:
Thanks. Yeah downloading the data with your script fixed it. Before I was getting around 9% Accuracy on the test haha. Thanks for the heads up about the cnn.

rcurtin:
Happy to help!  I'll go ahead and close the issue. :+1:

PR #236, opened 2024-09-04
Fix paths to test data

rcurtin:
While working on #235 I noticed that some of the paths for test data weren't correct after the restructuring of the repository.  This simple PR fixes that.


shrit (approved):
true, thanks for the PR,
It is hard to go all over it

PR #237, opened 2024-09-29
Regression

shrit:
Hi,

Ported a couple of examples of regression examples mostly linear

### DIFF
--- a/cpp/linear_regression/avocado_price_prediction/Makefile
+++ b/cpp/linear_regression/avocado_price_prediction/Makefile
@@ -1,0 +1,44 @@
+# This is a simple Makefile used to build the example source code.
+# This example might requires some modifications in order to work correctly on
+# your system.
+# If you're not using the Armadillo wrapper, replace `armadillo` with linker commands
+# for the BLAS and LAPACK libraries that you are using.
+
+TARGET := avocado_price_prediction
+SRC := avocado_price_prediction.cpp
+LIBS_NAME := armadillo
+
+CXX := g++
+CXXFLAGS += -std=c++17 -Wall -Wextra -O3 -DNDEBUG -fopenmp
+# Use these CXXFLAGS instead if you want to compile with debugging symbols and
+# without optimizations.
+# CXXFLAGS += -std=c++17 -Wall -Wextra -g -O0
+
+LDFLAGS  += -fopenmp
+# Add header directories for any includes that aren't on the
+# default compiler search path.
+INCLFLAGS := -I .
+# If you have mlpack or ensmallen installed somewhere nonstandard, uncomment and
+# update the lines below.
+# Uncomment the following if you are using the Scatter function for plotting
+# INCLFLAGS += -I/usr/include/python3.11
+# INCLFLAGS += -I/path/to/ensmallen/include/
+CXXFLAGS += $(INCLFLAGS)
+
+OBJS := $(SRC:.cpp=.o)
+LIBS := $(addprefix -l,$(LIBS_NAME))
+CLEAN_LIST := $(TARGET) $(OBJS)
+
+# default rule
+default: all
+
+$(TARGET): $(OBJS)
+	$(CXX) $(OBJS) -o $(TARGET) $(LDFLAGS) $(LIBS)
+
+.PHONY: all
+all: $(TARGET)
+
+.PHONY: clean
+clean:
+	@echo CLEAN $(CLEAN_LIST)
+	@rm -f $(CLEAN_LIST)
--- a/cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
+++ b/cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
@@ -2,140 +2,110 @@
 * Predicting Avocado's Average Price using Linear Regression
 * Our target is to predict the future price of avocados depending on various
 * features (Type, Region, Total Bags, ...).
-*
-* Dataset
-* 
-* Avocado Prices dataset has the following features:
-* PLU - Product Lookup Code in Hass avocado board.
-* Date - The date of the observation.
-* AveragePrice - Observed average price of single avocado.
-* Total Volume - Total number of avocado's sold.
-* 4046 - Total number of avocado's with PLU 4046 sold.
-* 4225 - Total number of avocado's with PLU 4225 sold.
-* 4770 - Total number of avocado's with PLU 4770 sold.
-* Total Bags = Small Bags + Large Bags + XLarge Bags.
-* Type - Conventional or organic.
-* Year - Year of observation.
-* Region - City or region of observation.
-*
 * Approach
 * 
-* In this example, first we will do EDA on the dataset to find correlation
-* between various features.
-* Then we'll be using onehot encoding to encode categorical features.
-* Finally we will use LinearRegression API from mlpack to learn the correlation
-* between various features and the target i.e AveragePrice.
+* In this example, we will be using one hot encoding to encode categorical
+* features. Then, we will use LinearRegression API from mlpack to learn
+* the correlation between various features and the target i.e AveragePrice.
 * After training the model, we will use it to do some predictions, followed by
 * various evaluation metrics to quantify how well our model behaves.
 */
+
 #include <mlpack.hpp>
 
 using namespace mlpack;
 
-//Drop the dataset header using sed, sed is a Unix utility that parses and transforms text."
-//!mkdir -p data && cat avocado.csv | sed 1d > avocado_trim.csv"
-//"Drop columns 1 and 2 (\"Unamed: 0\", \"Date\") as these are not required and their presence cause issues while loading the data."
-//!rm avocado_trim.csv"
-//"!mv avocado_trim2.csv avocado_trim.csv"
-
 int main()
 {
- //!wget -q https://datasets.mlpack.org/avocado.csv.gz"
- // Features 9 (Avocado type) and 11 (region of observation) are strings
- // (categorical), but armadillo matrices can contain only numeric information;
- // so, we have to explicitly define them as categorical in `datasetInfo`
- // this allows mlpack to map numeric values to each of those values,
- // which can later be unmapped to strings.
- // Load the dataset into armadillo matrix.
+  /** Dataset
+   * 
+   * Avocado Prices dataset has the following features:
+   * PLU - Product Lookup Code in Hass avocado board.
+   * Date - The date of the observation.
+   * AveragePrice - Observed average price of single avocado.
+   * Total Volume - Total number of avocado's sold.
+   * 4046 - Total number of avocado's with PLU 4046 sold.
+   * 4225 - Total number of avocado's with PLU 4225 sold.
+   * 4770 - Total number of avocado's with PLU 4770 sold.
+   * Total Bags = Small Bags + Large Bags + XLarge Bags.
+   * Type - Conventional or organic.
+   * Year - Year of observation.
+   * Region - City or region of observation.
+   *
+   * 9 Avocado type and 11 region of observation are categorical string,
+   * but armadillo matrices can contain only numeric information
+   * Therefore, we explicitly define them as categorical in `datasetInfo`
+   * this allows mlpack to map numeric values to each of those values,
+   * which can later be unmapped to strings.
+  */
+  /** PLEASE, delete the header of the dataset once you have downloaded the
+   * datset to your data/ directory. **/
+  // Load the dataset into armadillo matrix.
+  arma::mat matrix;
+  data::DatasetInfo info;
+  info.Type(9) = data::Datatype::categorical;
+  info.Type(11) = data::Datatype::categorical;
+  data::Load("../../../data/avocado.csv", matrix, info);
 
- arma::mat matrix;
- data::DatasetInfo info;
- info.Type(9) = data::Datatype::categorical;
- info.Type(11) = data::Datatype::categorical;
- data::Load("../../../data/avocado.csv", matrix, info);
- // Printing header for dataset.
- std::cout << std::setw(10) << "AveragePrice" << std::setw(14) 
-     << "Total Volume" << std::setw(9) << "4046" << std::setw(13) 
-     << "4225" << std::setw(13) << "4770" << std::setw(17) << "Total Bags" 
-     << std::setw(13) << "Small Bags" << std::setw(13) << "Large Bags" 
-     << std::setw(17) << "XLarge Bags" << std::setw(10) << "Type" 
-     << std::setw(10) << "Year" << std::setw(15) << "Region" <<  std::endl;
- std::cout << matrix.submat(0, 0, matrix.n_rows-1, 5).t() << std::endl;
- // Exploratory Data Analysis
- arma::mat output;
- data::OneHotEncoding(matrix, output, info);
- arma::Row<double> targets = arma::conv_to<arma::Row<double>>::from(output.row(0));
- // Labels are dropped from the originally loaded data to be used as features.
- output.shed_row(0);
+  arma::mat output;
+  data::OneHotEncoding(matrix, output, info);
+  arma::rowvec targets = arma::conv_to<arma::rowvec>::from(output.row(0));
+
+  // Labels are dropped from the originally loaded data to be used as features.
+  output.shed_row(0);
+  
+  // Train Test Split,
+  // The dataset has to be split into a training set and a test set. Here the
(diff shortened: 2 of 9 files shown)


cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+*/
+#include <mlpack.hpp>
+using namespace mlpack;
+using namespace mlpack::data;
+
+//Drop the dataset header using sed, sed is a Unix utility that parses and transforms text."
+//!mkdir -p data && cat avocado.csv | sed 1d > avocado_trim.csv"
+//"Drop columns 1 and 2 (\"Unamed: 0\", \"Date\") as these are not required and their presence cause issues while loading the data."
+//!rm avocado_trim.csv"
+//"!mv avocado_trim2.csv avocado_trim.csv"
```
rcurtin, line 39: Since this isn't a notebook, we don't have the ability to call out to the shell like this.  So maybe we either indicate in the comments that we expect the user to run these commands, or we adapt `download_data_set.py` to get the data into the right form first (or, the version we link to on datasets.mlpack.org is pre-adapted).

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+* Date - The date of the observation.
+* AveragePrice - Observed average price of single avocado.
+* Total Volume - Total number of avocado's sold.
+* 4046 - Total number of avocado's with PLU 4046 sold.
+* 4225 - Total number of avocado's with PLU 4225 sold.
+* 4770 - Total number of avocado's with PLU 4770 sold.
+* Total Bags = Small Bags + Large Bags + XLarge Bags.
+* Type - Conventional or organic.
+* Year - Year of observation.
+* Region - City or region of observation.
```
rcurtin, line 19: This could probably be formatted better and placed closer to where we actually load the data.  It seems a little out of place in the header comment, which should probably just have the basic idea of the file.

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+* between various features.
+* Then we'll be using onehot encoding to encode categorical features.
+* Finally we will use LinearRegression API from mlpack to learn the correlation
+* between various features and the target i.e AveragePrice.
+* After training the model, we will use it to do some predictions, followed by
+* various evaluation metrics to quantify how well our model behaves.
+*/
+#include <mlpack.hpp>
+using namespace mlpack;
+using namespace mlpack::data;
```
rcurtin, line 33:
```suggestion
#include <mlpack.hpp>

using namespace mlpack;
```

I don't think that `mlpack::data` is actually being used (I see calls to `data::DatasetInfo` later, for instance).

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+//Drop the dataset header using sed, sed is a Unix utility that parses and transforms text."
+//!mkdir -p data && cat avocado.csv | sed 1d > avocado_trim.csv"
+//"Drop columns 1 and 2 (\"Unamed: 0\", \"Date\") as these are not required and their presence cause issues while loading the data."
+//!rm avocado_trim.csv"
+//"!mv avocado_trim2.csv avocado_trim.csv"
+
+int main()
+{
+ //!wget -q https://datasets.mlpack.org/avocado.csv.gz"
```
rcurtin, line 43: This should probably be replaced with either a mention that the data needs to be downloaded with `download_data_set.py` or similar.  Whatever we do in the other examples should work fine.

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+ data::Load("../../../data/avocado.csv", matrix, info);
+ // Printing header for dataset.
+ std::cout << std::setw(10) << "AveragePrice" << std::setw(14) 
+     << "Total Volume" << std::setw(9) << "4046" << std::setw(13) 
+     << "4225" << std::setw(13) << "4770" << std::setw(17) << "Total Bags" 
+     << std::setw(13) << "Small Bags" << std::setw(13) << "Large Bags" 
+     << std::setw(17) << "XLarge Bags" << std::setw(10) << "Type" 
+     << std::setw(10) << "Year" << std::setw(15) << "Region" <<  std::endl;
+ std::cout << matrix.submat(0, 0, matrix.n_rows-1, 5).t() << std::endl;
+ // Exploratory Data Analysis
```
rcurtin, line 64: Should this comment still be here?

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+ arma::mat output;
+ data::OneHotEncoding(matrix, output, info);
+ arma::Row<double> targets = arma::conv_to<arma::Row<double>>::from(output.row(0));
+ // Labels are dropped from the originally loaded data to be used as features.
+ output.shed_row(0);
+ // Train Test Split,
```
rcurtin, line 70:
```suggestion
 output.shed_row(0);
 
 // Train Test Split,
```

It would probably help to add a little more whitespace throughout to split the sections logically.

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+ // The dataset has to be split into a training set and a test set. Here the
+ // dataset has 18249 observations and the `testRatio` is set to 20% of the
+ // total observations. This indicates the test set should have
+ // 20% * 18249 = 3649 observations and training test should have
+ // 14600 observations respectively.
+ arma::mat Xtrain;
+ arma::mat Xtest;
+ arma::Row<double> Ytrain;
+ arma::Row<double> Ytest;
```
rcurtin, line 79:
```suggestion
 arma::rowvec Ytrain;
 arma::rowvec Ytest;
```
Probably better to use the more common name here.

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+ data::Split(output, targets, Xtrain, Xtest, Ytrain, Ytest, 0.2);
+ // Convert armadillo Rows into rowvec. (Required by mlpacks' LinearRegression API in this format).
+ arma::rowvec yTrain = arma::conv_to<arma::rowvec>::from(Ytrain);
+ arma::rowvec yTest = arma::conv_to<arma::rowvec>::from(Ytest);
```
rcurtin, line 83: These shouldn't be necessary now... I'm not understanding why they were there before.

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+ LinearRegression lr(Xtrain, yTrain, 0.5);
+ arma::rowvec yPreds;
+ lr.Predict(Xtest, yPreds);
+ // Save the yTest and yPreds into csv for generating plots.
+ arma::mat preds;
+ preds.insert_rows(0, yTest);
+ preds.insert_rows(1, yPreds);
+ arma::mat histpreds = yTest - yPreds;
+ mlpack::data::Save("./data/predictions.csv", preds);
+ mlpack::data::Save("./data/predsDiff.csv", yPreds);
```
rcurtin, line 106: If we're `using namespace mlpack` then we can drop the `mlpack::` from here.

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+  * Create and train Linear Regression model.
+ */
+ LinearRegression lr(Xtrain, yTrain, 0.5);
+ arma::rowvec yPreds;
+ lr.Predict(Xtest, yPreds);
+ // Save the yTest and yPreds into csv for generating plots.
+ arma::mat preds;
+ preds.insert_rows(0, yTest);
+ preds.insert_rows(1, yPreds);
+ arma::mat histpreds = yTest - yPreds;
```
rcurtin, line 104: It looks like `histpreds` is not used anywhere?

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,140 @@
...
+ mlpack::data::Save("./data/predictions.csv", preds);
+ mlpack::data::Save("./data/predsDiff.csv", yPreds);
+ /*
+  * Model Evaluation,
+  * Test data is visualized with `yTest` and `yPreds`, the blue points
```
rcurtin, line 109: But this program can't visualize anything, it can only print things.  There are no plots.

rcurtin (commented):
I only reviewed one example so far.  I'm not entirely sure that all of these notebooks will adapt well to C++-only programs.  Many of them focus on visualization and exploration, which isn't really a thing we can do just in a standalone program.  So I think either they need to be really carefully gone over and all irrelevant parts removed, or maybe we leave them as notebook-only examples.  Let me know what you think.

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
shrit: exporting is the most hideous part, I forget to check these comments, I will go over them

shrit:
I agree, I am trying to have at least one usage from the examples at least per method, because we have so many methods and really few examples.

I know if someone looks at the docs they would find some examples from them, but I agree, no need anymore to import the old ones, we can be more creative and have new ones.

Let us merge these three once I have solved the comments, I have tested them locally and I know they work, and they are old, as I did them one month ago or even more

rcurtin:
> Let us merge these three once I have solved the comments, I have tested them locally and I know they work, and they are old, as I did them one month ago or even more

Sounds good to me---my primary concern is just making sure that all the comments in the file next to the code (which is what users will actually be looking at) make sense and don't reference things specific to the notebook but not in this version. :+1:

rcurtin:
Also, looks like the steps in `.ci/` just need slight modification to also install Pandas.

cpp/linear_regression/avocado_price_prediction/Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+OBJS := $(SRC:.cpp=.o)
+LIBS := $(addprefix -l,$(LIBS_NAME))
+CLEAN_LIST := $(TARGET) $(OBJS)
+
+# default rule
+default: all
+
+$(TARGET): $(OBJS)
+	$(CXX) $(OBJS) -o $(TARGET) $(LDFLAGS) $(LIBS)
```
rcurtin, line 36: Do we need to get `CXXFLAGS` in here too?

cpp/linear_regression/avocado_price_prediction/avocado_price_prediction.cpp
```diff
@@ -0,0 +1,111 @@
...
+   * Region - City or region of observation.
+   *
+   * 9 Avocado type and 11 region of observation are categorical string,
+   * but armadillo matrices can contain only numeric information
+   * Therefore, we explicitly define them as categorical in `datasetInfo`
+   * this allows mlpack to map numeric values to each of those values,
+   * which can later be unmapped to strings.
+  */
+  /** PLEASE, delete the header of the dataset once you have downloaded the
+   * datset to your data/ directory. **/
```
rcurtin, line 42: Definitely we should support headers soon, but for now should we modify the dataset downloader to strip the header?

cpp/linear_regression/california_housing_price_prediction/Makefile
```diff
@@ -0,0 +1,44 @@
...
+CXXFLAGS += $(INCLFLAGS)
+
+OBJS := $(SRC:.cpp=.o)
+LIBS := $(addprefix -l,$(LIBS_NAME))
+CLEAN_LIST := $(TARGET) $(OBJS)
+
+# default rule
+default: all
+
+$(TARGET): $(OBJS)
```
rcurtin, line 35: Same here with `CXXFLAGS`

cpp/linear_regression/california_housing_price_prediction/california_house_price_prediction.cpp
```diff
@@ -0,0 +1,138 @@
+/**
+* Predicting  California House Prices with  Linear Regression
```
rcurtin, line 2: Code-wise, this is basically the same example as the avocado one, but with a different dataset. Do you think it is worth including? I think quality over quantity is better, so if this is not demonstrating some new or different functionality vs. the avocado example, I would vote to only keep one of the two.

cpp/linear_regression/salary_prediction/Makefile
```diff
@@ -0,0 +1,44 @@
+# This is a simple Makefile used to build the example source code.
+# This example might requires some modifications in order to work correctly on
+# your system.
+# If you're not using the Armadillo wrapper, replace `armadillo` with linker commands
+# for the BLAS and LAPACK libraries that you are using.
+
+TARGET := salary_prediction 
+SRC := salary-prediction.cpp
```
rcurtin, line 8: We should probably rename this to `salary_prediction.cpp` for consistency.

scripts/download_data_set.py
```diff
@@ -164,6 +170,20 @@ def covertype_dataset():
...
+    progress_bar("housing.csv", california)
+
+def avocado_dataset():
+    print("Downloading the avocado price prediction dataset...")
+    avocado = requests.get("https://datasets.mlpack.org/avocado.csv.gz")
+    progress_bar("avocado.csv.gz", avocado)
+    ungzip("avocado.csv.gz", "avocado.csv")
+    avocado_data = pull_csv("avocado.csv")
+    avocado_data = avocado_data.iloc[:, 2:]
+    avocado_data.to_csv("avocado.csv", index=False)
```
rcurtin, line 185: Or are you already stripping the header here? If so, it would probably be better to omit the comments from the files and work on adding header support to `data::Load()` quickly, then we don't have to come back to the comments here.

scripts/download_data_set.py
```diff
@@ -210,18 +230,20 @@ def all_datasets():
...
         color: will download the dominant color dataset
         covertype: will download the forest covertype dataset
+        electricity : will download electricty_consumption_dataset
+        iris : will downlaod the iris dataset
+        loan: will download the loan default dataset
+        mnist : will download mnist dataset
+        pima: will download the pima diabetes dataset
+        salary: will download the salary dataset
+        spam : will download the spam dataset
+        stock : will download stock_exchange dataset
```
rcurtin, line 246: Thanks, nice to fix all this!

cpp/linear_regression/salary_prediction/salary-prediction.cpp
```diff
@@ -0,0 +1,96 @@
...
+ * Objective:
+ * We have to predict the salary of an employee given how many years of experience they have.
+ * Approach:
+ * So in this example, we will train a Linear Regression model to learn the
+ * correlation between the number of years of experience of each employee and
+ * their respective salary.
+ * Once the model is trained, we will be able to do some sample predictions.
+*/
+#include <mlpack.hpp>
+#include <cmath>
```
rcurtin, line 12: I think this include is unnecessary.

rcurtin (commented):
Looks good overall. In the longer term I am a little worried that we will have too many similar examples that will cause users to get lost, but I think we can sort that out later.  I left a comment to that effect in the review, but up to you how you want to handle it (and if the answer is let's merge all three for now and sort it out later, that's fine with me).

scripts/download_data_set.py
shrit: we need the header support for `data::Load`

rcurtin: Totally agree, but what I mean in this comment is, it looks like you are stripping the header from the avocado dataset and also other datasets.  So I think we can remove the warning comments in the code that say 'make sure the header is removed', because it is already removed here---and also because we will be adding header removal support very soon to `data::Load()`.

PR #241, opened 2025-01-09
Fix embedded cmake

shrit:


embedded/crosscompile_random_forest/CMake/crosscompile-arch-config.cmake
```diff
@@ -19,77 +19,82 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -flto")
...
+
+if (BOARD_NAME)
+  set(ARCH_NAME "${BOARD_NAME}")
+endif()
+
+string(TOUPPER ${ARCH_NAME} ARCH)
 
 # Set specific platforms CMAKE CXX flags.
-if(BOARD MATCHES "RPI0" OR BOARD MATCHES "RPI1" OR BOARD MATCHES "ARM11")
+if(ARCH MATCHES "RPI0" OR ARCH MATCHES "RPI1" OR ARCH MATCHES "ARM11")
```
rcurtin, line 37: These should be `STREQUAL` not `MATCHES`; I found this is another branch I am working on.  When it's `MATCHES` but you specify `CORTEXA76`, the first thing it matches is line 42 (`CORTEXA7`) and the wrong compiler options get used.

embedded/embedded_cmake_template/CMake/crosscompile-arch-config.cmake
```diff
@@ -19,77 +19,82 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -flto")
...
+
+if (BOARD_NAME)
+  set(ARCH_NAME "${BOARD_NAME}")
+endif()
+
+string(TOUPPER ${ARCH_NAME} ARCH)
 
 # Set specific platforms CMAKE CXX flags.
-if(BOARD MATCHES "RPI0" OR BOARD MATCHES "RPI1" OR BOARD MATCHES "ARM11")
+if(ARCH MATCHES "RPI0" OR ARCH MATCHES "RPI1" OR ARCH MATCHES "ARM11")
```
rcurtin, line 37: Same here, should be `STREQUAL`.

embedded/embedded_cmake_template/main.cpp
```diff
@@ -0,0 +1,23 @@
...
+ * @author Omar Shrit
+ */
+
+#include <mlpack.hpp>
+
+using namespace mlpack;
+
+int main(int argc, char** argv)
+{
+  Log::Info << "Welcome to an mlpack crosscompiled binary." <<std::endl;
```
rcurtin, line 22:
```suggestion
  Log::Info << "Welcome to an mlpack crosscompiled binary." << std::endl;
```
I know, super pedantic 😄

rcurtin (approved):
Looks good to me, just a few comments.  For the STREQUAL/MATCHES comments, you can handle them here if you want, or we can fix them upstream in mlpack then update the files back here.  Up to you. :+1:

PR #247, opened 2025-04-04
Include download of loan default dataset when downloading 'all' datasets.

derekeverett:
The loan default dataset was the only dataset that is currently excluded, and can lead to confusion when following current documentation to run the cpp/decision_tree/loan_default_prediction example.


rcurtin (approved):
Thanks @derekeverett!  This one got overlooked, I guess.

rcurtin:
(Looks like the style check job has issues of its own.  But I'm not going to worry about that here.)

